PR #134, opened 2023-06-16
Fix export: skip projection and fcr phenotype rows if none

Anto8819:
Also pushing export relationship for destination and via type.


zsinnema (approved):
lgtm

PR #137, opened 2023-08-21
SCKAN-145 feat: Allow check for duplicates with just one AnatomicalEntity set

afonsobspinto:
closes https://metacell.atlassian.net/browse/SCKAN-145

- Updates UI to allow searches with either (or both) source or destination entities set.

https://github.com/MetaCell/sckan-composer/assets/19196034/69558475-8118-44b5-b819-99b903b034cd


zsinnema (approved):
lgtm

PR #142, opened 2023-09-06
SCKAN-148 feat: Add move forward feature on triage step

afonsobspinto:
closes https://metacell.atlassian.net/browse/SCKAN-148

- Adds new filter on sentence serializer to exclude some sentence ids from the search
- Generates api
- Updates sentence details onClick handler to search for the next open sentence that matches the previously used search params excluding itself and navigates to the results page if exists or to the home page if it doesn't

https://github.com/MetaCell/sckan-composer/assets/19196034/fb0f3cbb-dc6c-44f7-8370-ffc9bfd875be


ddelpiano (dismissed):

zsinnema (approved):
@ddelpiano we might want to add an indicator / message / flash to visually show the user that "something happened" and that he/she moved to the next record

PR #171, opened 2023-11-02
SCKAN-181 Clearing search using 'X' does not work

Aiga115:
Issue #SCKAN-181
Problem: Clearing search using 'X' does not work
Solution: 
Autocomplete mui has a clear icon to deselect all the items, earlier it was handled by custom icon (that is being rendered with the help of endIcon property), which doesn't do anything onclick, while the autocomplete's own clear icon was disabled by disableClearable property. 
I deleted this property and deleted endIcon, which in my opinion was not useful at all, since it does even appear when there are no selected items (because it used to appear anytime when the text filed was on focused state)

the video of solution

https://github.com/MetaCell/sckan-composer/assets/67194168/98b5811f-ed10-49cc-844c-c205e89d3a08


zsinnema:
@Aiga115 for me this doesn't work? Can you please check again?
BTW can you first please pull latest? I've merged develop
thanks

zsinnema (approved):

PR #174, opened 2023-11-15
#198 fix loading problem, hide drag and drop for destinations

Salam-Dalloul:
fix loading problem
hide drag and drop for destinations


zsinnema (approved):

zsinnema:
@Salam-Dalloul can you again have a look , still some merge conflicts

PR #175, opened 2023-11-15
#197 Add current options as part of selected options

Salam-Dalloul:


zsinnema (approved):

frontend/src/components/Forms/StatementForm.tsx
```diff
@@ -1,419 +1,529 @@
...
-                    CustomFooter: CustomFooter,
-                },
-            },
+          label: "From",
+          placeholder: "Look for connections",
+          searchPlaceholder: "Search for connections",
+          noResultReason: "No prior connections found",
+          disabledReason: "",
+          onSearch: async (searchValue: string, formId: string) => {
+            return getAnatomicalEntitiesForForm(
```
afonsobspinto, line 264: This doesn't seem right. It should be using `searchFromEntitiesVia`

afonsobspinto (changes requested):
The search on the from is different from the search on the anatomical entities but it's being treated equally in the current PR. Please keep the logic from 'searchFromEntitiesVia' and 'searchFromEntitiesDestination'

frontend/src/services/CustomDropdownService.ts
```diff
@@ -1,217 +1,288 @@
...
+  const viaIndex = getFirstNumberFromString(formId);
+  if (
+    viaIndex == null ||
+    statement.vias == null ||
+    statement.vias[viaIndex] == null
+  ) {
+    return [];
+  }
+  const viaOrder = statement.vias[viaIndex].order;
+  const anatomicalEntities = getEntitiesBeforeOrder(statement, viaOrder);
```
afonsobspinto, line 128: Although practically speaking it's unlikely that we will need it, we should still use the current selected options in the searchFromEntitiesX functions to populate the anatomicalEntities options (without creating duplicates)

afonsobspinto (changes requested):
Looks better but see my comment above :arrow_up:

afonsobspinto (approved):

zsinnema:
@Salam-Dalloul can you have a look at the conflicts? Thanks!

zsinnema:
@Salam-Dalloul still a conflict remains. Can you have a look? thanks!

PR #178, opened 2023-11-16
#205 grouping in the dropdown search opened by the From option

Salam-Dalloul:


frontend/src/helpers/dropdownMappers.ts
```diff
@@ -94,4 +95,8 @@ export function sortFromViasEntities(entities: Option[]){
     // Origins comes last
     return groupA === 'origins' ? 1 : -1;
   });
+}
+
+export function getViasGroupLabel(currentIndex: number | null) {
```
afonsobspinto, line 100: If it can return the label for the origins I would give this method a different name: getGroupLabel only instead of getViasGroupLabel maybe?

frontend/src/components/Forms/StatementForm.tsx
```diff
@@ -289,8 +283,11 @@ const StatementForm = (props: any) => {
             refreshStatement();
           },
           errors: "",
-          mapValueToOption: (anatomicalEntities: any[]) =>
-            mapAnatomicalEntitiesToOptions(anatomicalEntities, ViasGroupLabel),
+          mapValueToOption: (anatomicalEntities: any[], formId: any) => {
+            const currentIndex = getFirstNumberFromString(formId);
+            const label = getViasGroupLabel(currentIndex?? - 1)
```
afonsobspinto, line 289: This is assuming that the selected values are coming **always** from the via immediately above the current one. That's not correct, but I understand the struggle.

Without changing the backend model, we need to implement a new function that looks for each anatomical entity selected in the statement.origins and statement.vias, preferably in this order. When it finds it, you know to which layer the entity belongs to.

Origins  = layer 0
Via 0 = layer 1 (order + 1)
Via 1 = layer 2,
etc...

If there are anatomical entities duplicated, for now we assume it connects to the first you can find
If you don't find the anatomical entity in any of the origins or vias, the values is invalid (should never happen but if it does the UI should give some sort of feedback, not sure if figma addresses this)

afonsobspinto (changes requested):

frontend/src/components/Forms/StatementForm.tsx
afonsobspinto: For complete transparency, this method might need to change later on, depending on the customer reply to Dario's question. If it takes you too much time to address this, it might make sense to wait for the reply before proceeding. Tagging @ddelpiano here so he can be aware and provide guidance

Salam-Dalloul: @afonsobspinto @ddelpiano i have pushed a fix for this grouping issue, based on what @afonsobspinto mentioned,  it's no longer depends on the index,  can you pleas check the latest pushed code!

afonsobspinto (approved):
Edit:

We already got a reply so I would suggest to apply the following in this PR:
image

____


Looks good to me.
~~We are probably going to treat the duplicates slightly different from the current behavior but we can address that in other PR imo.~~

afonsobspinto (approved):

PR #179, opened 2023-11-23
Feature/sckan 207

Salam-Dalloul:


frontend/src/components/Widgets/CustomEntitiesDropdown.tsx
```diff
@@ -324,41 +318,38 @@ export default function CustomEntitiesDropdown({
...
-      } finally {
-        setIsLoading(false);
-      }
-    };
+  const fetchData = useCallback(async () => {
+    try {
+      const options = await onSearch(inputValue, id, selectedOptions);
+      const allOptions = [...selectedOptions, ...options];
+      const sortedOptions =
+        selectType && selectType === "from_entities"
```
afonsobspinto, line 333: This check on selectType is not nice for a generic component. 
I would suggest to have a new prop instead. Something like 'postProcessOptions' or 'sortOptions' what do you think?

afonsobspinto (changes requested):
Even tho changing the via/destination type is effectively updating the entity in the database, changing tabs seems to rely in the old entity content:

https://github.com/MetaCell/sckan-composer/assets/19196034/1a4e633e-f926-4cdf-ad4c-04b19039c2a7

PR #189, opened 2023-12-11
Feature/sckan 196

afonsobspinto:
This PR closes the first iteration of the compatibility with graphs.
Most of the frontend work was already reviewed so I would focus this PR review in the backend changes (but feel free to review everything :smile: ) in the context of https://metacell.atlassian.net/browse/SCKAN-173


zsinnema (changes requested):

backend/composer/api/serializers.py
```diff
@@ -154,23 +155,159 @@ class Meta:
...
+            "type",
+            "anatomical_entities",
+            "from_entities"
+        )
+
+    def to_representation(self, instance):
+        """
+        Custom representation for Via.
+        """
+        representation = super().to_representation(instance)
```
zsinnema, line 184: @afonsobspinto I suggest to use a methodfield here to compute the previous vias
see https://www.django-rest-framework.org/api-guide/fields/#serializermethodfield

zsinnema (changes requested):

backend/composer/api/serializers.py
```diff
@@ -154,23 +155,159 @@ class Meta:
...
+                    order=current_order - 1
+                ).first()
+
+                if previous_via:
+                    # Serialize anatomical_entities from previous via
+                    previous_entities = AnatomicalEntitySerializer(
+                        previous_via.anatomical_entities.all(),
+                        many=True
+                    ).data
+                    representation['from_entities'] = previous_entities
```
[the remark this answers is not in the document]
zsinnema: see my comment

backend/composer/services/graph_service.py
```diff
@@ -0,0 +1,128 @@
+JOURNEY_DELIMITER = '\\'
+
+
+def generate_paths(origins, vias, destinations):
+    paths = []
+    # Calculate the total number of layers, including origins and destinations
+    number_of_layers = len(vias) + 2 if vias else 2
+
+    # Handle direct connections from origins to destinations
+    for origin in origins:
```
zsinnema, line 10: @afonsobspinto this is very "3rd generation" development
can't we use the database to retrieve this and remove the for loops ?

zsinnema (changes requested):

backend/composer/services/graph_service.py
afonsobspinto: Without changing the model to contain information about which **layer** we are connecting to (or from) I don't think we can get rid of these for loops unless we replace them with custom sql queries, but I'm not sure that's what you are aiming for here. Would you mind to elaborate a bit on what's your idea @zsinnema ?

afonsobspinto: The core problem / complexity of this service is that we don't have an exact graph information. 
We know that a node is connected to some entity from a previous layer but not exactly to each one. 

Roughly, in order to beat this problem, I'm going layer by layer and constructing the paths as we go.

afonsobspinto: The second part of the algorithm takes all those paths found and 'consolidates' them.

zsinnema: @afonsobspinto currently I have no hours left to work on this. My suggestion would be to retrieve all "data" at once and then use python to structure it. Please check the debug toolbar queries when retrieving the connectivity statements. It will show you that this request is executing hundreds/thousands queries. The goal here would be to reduce the number of queries for retrieving CS's

afonsobspinto: I understand [7ce58de](https://github.com/MetaCell/sckan-composer/pull/189/commits/7ce58de75f731129c70221e84e46440e4b9204c6) should solve the many database queries.

backend/composer/models.py
```diff
@@ -522,7 +522,7 @@ def save(self, *args, **kwargs):
             self.reference_uri = create_reference_uri(self.pk)
             self.save(update_fields=["reference_uri"])
 
-    def update_origins(self, origin_ids):
+    def set_origins(self, origin_ids):
         self.origins.clear()
```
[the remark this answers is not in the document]
afonsobspinto: [5599752](https://github.com/MetaCell/sckan-composer/pull/189/commits/55997521bef1237782edf75e9763492ed91688a1)

backend/composer/api/serializers.py
```diff
@@ -405,6 +532,14 @@ def to_representation(self, instance):
             ).data
         return representation
 
+    def update(self, instance, validated_data):
+        # Remove 'vias' and 'destinations' from validated_data if they exist
+        validated_data.pop('via_set', None)
+        validated_data.pop('destinations', None)
```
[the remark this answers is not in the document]
afonsobspinto: Marking the fields as read_only also removes them from the automatic schema generation which breaks the current frontend implementation

backend/composer/api/serializers.py
afonsobspinto: Cool suggestion. Check:
[c47bbc3](https://github.com/MetaCell/sckan-composer/pull/189/commits/c47bbc3ec674f6140a8965953ae5e16e7732e06c)

backend/composer/services/state_services.py
```diff
@@ -121,9 +121,16 @@ def compile_journey(connectivity_statement) -> List[str]:
            A string with each journey path description on a new line.
        """
         # Extract origins, vias, and destinations from the connectivity statement
+        Via = apps.get_model('composer', 'Via')
```
zsinnema, line 124: @afonsobspinto imo we don't need this, we can use the queryset `connectivity_statement.via_set` 
and then add the prefetch related and/or select_related (if needed)

however my biggest concern here is not the prefetch/select but the many queries we do in all the loops
suggestion: query all needed data and store this in a Python "graph" (list with dicts or dict with lists and use this object to compile the path

backend/composer/api/serializers.py
```diff
@@ -465,7 +465,20 @@ def get_available_transitions(self, instance) -> list[CSState]:
     def get_has_notes(self, instance) -> bool:
         return instance.has_notes
 
-    def get_statement_preview(self, instance) -> str:
+    def get_journey(self, instance):
```
zsinnema, line 468: @afonsobspinto this works but I suggest next time to create a separate serializer for the list view. For example we can create a "base" CS serializer without the journey and preview fields. Then use (extend) this one for the list serializer (no extra fields) and extend it (with journey and preview fields) for the get view. Then in the views.py we can pick the serializer per method (list or get). This way it's more clear what the "interface" per api is. For now we can leave it as is.

backend/composer/api/serializers.py
afonsobspinto: I reverted this commit, because the serializer was not including the from_entities in the schema

backend/composer/services/state_services.py
afonsobspinto: With use of prefetch_related there are no database queries done in the generate_paths method. I'm asserting this [here](https://github.com/MetaCell/sckan-composer/blob/feature/SCKAN-196/backend/tests/test_journey.py#L48)

backend/composer/api/serializers.py
afonsobspinto: @ddelpiano can you create a tech debt card for this? It's both cleaner and might help a bit in the performance if the calculation of the additional fields is 'expensive'

afonsobspinto: [d4fd9f9](https://github.com/MetaCell/sckan-composer/pull/189/commits/d4fd9f9c210c3536eb835774d44ea18965f8cd34)

afonsobspinto: This actually solves the performance issue on the list view as well. There's probably some attribute besides the journey that is slow to compute. Maybe the from_entities of both vias and destinations

zsinnema (approved):

PR #202, opened 2024-01-04
Feature/sckan 226

afonsobspinto:


backend/composer/services/cs_ingestion/neurondm_script.py
```diff
@@ -171,31 +268,23 @@ def main(local=False, anatomical_entities=False, anatent_simple=False):
...
-    myFile.close()
-
-   # breakpoint()
-    return fcs
+    neurons = config.neurons()
+
+    fcs = [for_composer(n) for n in neurons]
+    composer_statements = [item for item in fcs if item is not None]
+
+    # TODO: Confirm if this can be deleted
```
afonsobspinto, line 276: Previously we were just getting primitives (strings) in the dict produced from the neurondm data. I changed it to use some complex objects to facilitate the ingestion command, due to that I'm not sure about what we want to show in the csv that was being produced, for those objects. I commented out the csv creation for now but I would like confirmation if we don't need this. If we need it, then what would you like it to show?

backend/composer/models.py
```diff
@@ -220,7 +220,7 @@ class AnatomicalEntity(models.Model):
     """Anatomical Entity"""
 
     name = models.CharField(max_length=200, db_index=True)
-    ontology_uri = models.URLField()
+    ontology_uri = models.URLField()  # TODO: Confirm it is acceptable to have multiple anatomical entities with the same uri.
```
afonsobspinto, line 223: I noticed that we have multiple anatomical entities for the same ontology_uri. Is that expected? When transforming the uris from neurondm into anatomical entities in sckan I'm getting the first match

backend/composer/services/cs_ingestion/cs_ingestion_services.py
```diff
@@ -40,253 +39,351 @@
...
+def generate_connectivity_statement_defaults(statement: Dict, sentence: Sentence) -> Dict:
+    return {
+        "sentence": sentence,
+        "knowledge_statement": statement[LABEL],
+        "sex": get_sex(statement),
+        "circuit_type": get_circuit_type(statement),
+        "functional_circuit_role": get_functional_circuit_role(statement),
+        "phenotype": get_phenotype(statement),
+        "projection_phenotype": get_projection_phenotype(statement),
+        "state": CSState.EXPORTED,  # TODO: Confirm if this is fine
```
afonsobspinto, line 190: Previously the state change was running transitions instead of starting the state in the desired place already.
Because the previous logic needed updated I opted for this simple approach but I'm not sure if I'm missing some context

backend/composer/services/cs_ingestion/neurondm_script.py
```diff
@@ -88,42 +85,142 @@ def for_composer(n, cull=False):
...
+    uri_type_dict = {}
+    for predicate, type_name in predicate_type_map.items():
+        for uri in lpes_func(predicate):
+            uri_type_dict[uri] = type_name
+    return uri_type_dict
+
+
+# FIXME: The following algorithm will behave incorrectly for cases:
+#  where an anatomical entity has more than one 'role' (Origin, Via, Destinaion)
+#  where we have connections that jump layers
```
afonsobspinto, line 120: I can't improve these points until I see how neurondm models these edge cases @ddelpiano

backend/composer/services/cs_ingestion/neurondm_script.py
```diff
@@ -88,42 +85,142 @@ def for_composer(n, cull=False):
...
+                depth = 0
+            elif primary_uri in expected_vias or secondary_uri in expected_vias:
+                matched_uri = primary_uri if primary_uri in expected_vias else secondary_uri
+                result['vias'].append(Via({matched_uri}, from_entities, depth, expected_vias.get(matched_uri)))
+                depth += 1
+            elif primary_uri in expected_destinations or secondary_uri in expected_destinations:
+                matched_uri = primary_uri if primary_uri in expected_destinations else secondary_uri
+                result['destinations'].append(Destination({matched_uri}, from_entities, expected_destinations.get(matched_uri)))
+            else:
+                raise Exception(f"{current_entity} not found in axiom.")
```
afonsobspinto, line 150: This should also be reported to the neurondm team @ddelpiano

backend/composer/services/cs_ingestion/cs_ingestion_services.py
```diff
@@ -40,253 +39,351 @@
...
+                add_ingestion_system_note(connectivity_statement)
 
-def pick_first_phenotype(phenotypes_list, statement):
-    if len(phenotypes_list) == 0:
-        return None
-    elif len(phenotypes_list) > 1:
-        logging.warning(
-            f'Multiple phenotypes found, ignore the following {phenotypes_list[1:]} at statement {statement[LABEL]}')
-    return phenotypes_list[0]
+    update_forward_connections(valid_statements)
```
afonsobspinto, line 58: We do this after all the statements are ingested because there could be references in the forward connection to statements to be ingested in the current batch

backend/composer/services/cs_ingestion/cs_ingestion_services.py
```diff
@@ -40,253 +39,351 @@
...
-                logging.warning(f'Skip phenotype {p} at statement {statement[LABEL]} not found in composer db ')
-    return valid_phenotypes
+        connectivity_statement, statement_created = get_or_create_connectivity_statement(statement, sentence)
 
+        if not statement_created:
+            if should_overwrite(connectivity_statement, statement):
+                defaults = generate_connectivity_statement_defaults(statement, sentence)
+                ConnectivityStatement.objects.filter(id=connectivity_statement.id).update(**defaults)
+                update_many_to_many_fields(connectivity_statement, statement)
+                add_ingestion_system_note(connectivity_statement)
```
afonsobspinto, line 56: This adds a new note to inform the users about an overwrite that happened due to the ingestion in progress

backend/composer/models.py
zsinnema: @afonsobspinto sounds good to me. In case the user would like to explicit pick sort the AE we would need to add a sort order attribute to AE. And then we would need to sort here by the sort order. 

an AE can have an alias, the alias is in Composer an AE with a different name but the same ontology_uri

backend/composer/services/cs_ingestion/cs_ingestion_services.py
zsinnema: @afonsobspinto I'm not a big fan of changing the records via the `.update` method
please have a look here: https://docs.djangoproject.com/en/5.0/topics/db/queries/#updating-multiple-objects-at-once
fourth alinea the `Beware part`

better use the `.update_or_create(filter,filter,defaults=defaults`

zsinnema: the latter will trigger the save methods and signals (that are connected to the model)

backend/composer/services/cs_ingestion/cs_ingestion_services.py
zsinnema: question: @afonsobspinto what will happen in case the ingestion of a specific statement (let's call it StatA) was rejected and another statement (StatB) has a forward connection to StatA

imo StatB should also not be ingested because of data lost

May be the whole ingestion is an atomic operation: only do it if everything can be processed

cc: @ddelpiano

backend/composer/services/cs_ingestion/cs_ingestion_services.py
zsinnema: @afonsobspinto see my previous comment about the `.update` method

backend/composer/services/cs_ingestion/neurondm_script.py
zsinnema: @afonsobspinto @ddelpiano in case we ingest parts of the neuromdm neurons I propose to only ingest the neurons that can be 100% ingested, if a neuron has dependencies or depends on other neurons that couldn't be ingested then we should not ingest the neuron

backend/composer/services/cs_ingestion/neurondm_script.py
zsinnema: @afonsobspinto imo we can remove the csv creation. I'm not aware of anything using this csv, @ddelpiano do you?

zsinnema (changes requested):

backend/composer/services/cs_ingestion/neurondm_script.py
afonsobspinto: I kept it per @ddelpiano request but it will only show the uri and label of what was successfully ingested

backend/composer/services/cs_ingestion/neurondm_script.py
afonsobspinto: I'm now doing a double validation. I check if the statements are valid on their own properties, and then I check if they are valid on their dependencies (the forward connections).
 
This way, if a valid on their own statement refers an invalid statement in their forward connection, the valid on their own statement will not be ingested. This goes against what Sue mentioned in the shared channel but I trust that this was mentioned to her posterior to that post of hers.

The second check relies on the idea that there are no partial ingestions.

backend/composer/services/cs_ingestion/cs_ingestion_services.py
afonsobspinto: I addressed the first part of the question [here](https://github.com/MetaCell/sckan-composer/pull/202#discussion_r1450993973).

I also wrapped the whole ingestion in a transaction

backend/composer/services/cs_ingestion/cs_ingestion_services.py
afonsobspinto: Thank you django specialist @zsinnema 😅

backend/composer/models.py
```diff
@@ -500,6 +500,15 @@ def npo_approved(self, *args, **kwargs):
...
 
+    @transition(
+        field=state,
+        source=CSState.DRAFT,
+        target=CSState.EXPORTED,
+        permission=lambda instance, user: user.username == "system",
+    )
+    def exported_from_ingestion(self, *args, **kwargs):
+        ...
+
```
afonsobspinto, line 511: @zsinnema please have a look at this addition. I'm not sure if it's the best way to do what I want which is to move the newly ingested statements that are on draft by default to exported

zsinnema: @afonsobspinto I would not add this transition since it will become part of the "statement flow" process.

The way you solved it last time with the `.update` sound better to me. What do you think?

afonsobspinto: I'm afraid my original strategy doesn't work because on updates we get an error 'Direct state modification is not allowed' thus the need for the new transition. But I do understand your concerns. Do you have any alternative solution @zsinnema ?

backend/composer/services/cs_ingestion/cs_ingestion_services.py
```diff
@@ -180,37 +180,33 @@ def get_or_create_sentence(statement: Dict) -> Tuple[Sentence, bool]:
...
+        "sex": get_sex(statement),
+        "circuit_type": get_circuit_type(statement),
+        "functional_circuit_role": get_functional_circuit_role(statement),
+        "phenotype": get_phenotype(statement),
+        "projection_phenotype": get_projection_phenotype(statement),
+        "reference_uri": statement[ID],
+        "state": CSState.EXPORTED,
+    }
+
+    if ConnectivityStatement.objects.filter(reference_uri__exact=reference_uri).exists():
```
zsinnema, line 195: @afonsobspinto suggestion: did you consider using `.get_or_create` (see https://docs.djangoproject.com/en/5.0/ref/models/querysets/#get-or-create) ?

backend/composer/services/cs_ingestion/neurondm_script.py
```diff
@@ -31,55 +30,35 @@ def lrdf(neuron, predicate):
...
+        label=str(n.origLabel),
+        origins=origins,
+        destinations=destinations,
+        vias=vias,
+        species=lpes(n, ilxtr.hasInstanceInTaxon),
+        sex=lpes(n, ilxtr.hasBiologicalSex),
+        circuit_type=lpes(n, ilxtr.hasCircuitRolePhenotype),
+        circuit_role=lpes(n, ilxtr.hasFunctionalCircuitRolePhenotype),
+        phenotype=lpes(n, ilxtr.hasAnatomicalSystemPhenotype),
+        #classification_phenotype=lpes(n, ilxtr.hasClassificationPhenotype),
```
afonsobspinto, line 54: I noticed that we were not using this in composer so I commented it out for now. Please let me know if we should be using this somewhere in composer

backend/composer/services/cs_ingestion/cs_ingestion_services.py
```diff
@@ -1,292 +1,455 @@
...
+    if len(current_vias) != len(new_vias):
+        return True
+
+    for current_via, new_via in zip(current_vias, new_vias):
+        if (new_via.anatomical_entities != current_via['anatomical_entities'] or
+                new_via.from_entities != current_via['from_entities']):
+            return True
+
+    # Check for changes in destinations
+    # Fixme: We may need to change this algorithm when multi-destination is supported by neurondm
```
afonsobspinto, line 267: For the vias we can compare the vias with the same order and we are good, but for destinations there's no way to pair destinations together. Right now, we should always have just 1 destination coming from neurondm but if in the future this changes the algorithm presented here might not be the most scientifically accurate

afonsobspinto:
While testing this I noticed that Notes are created in CET timezone but the frontend is displaying it in my system timezone:
image
cc @ddelpiano

afonsobspinto:
@zsinnema can you re-review this without considering what we are still discussing in https://metacell.atlassian.net/browse/SCKAN-248? 
@ddelpiano asked me to have https://metacell.atlassian.net/browse/SCKAN-248 done in 2 PR's so that we can unblock @D-GopalKrishna and @Salam-Dalloul sooner rather than later

backend/composer/services/cs_ingestion/cs_ingestion_services.py
```diff
@@ -1,292 +1,522 @@
...
+                # Build the reason string
+                current_reason = (f"statement with id {backward_cs.id} is invalid because its "
+                                  f"forward connection with id {connectivity_statement.id} is invalid")
+                if previous_reason:
+                    current_reason += f" because {previous_reason}"
+
+                # Transition the backward statement to an invalid state with the built reason
+                do_transition_to_invalid(backward_cs, current_reason)
+            else:
+                # TODO: Should we update the reason?
```
afonsobspinto, line 519: If a statement is already invalid per se and should also be marked as invalid due to a invalid forward connection, should we just have 1 note for the per se or add a new note for the forward connection as well?

backend/composer/services/cs_ingestion/cs_ingestion_services.py
```diff
@@ -1,292 +1,523 @@
...
+            return True
+
+    # Check for changes in destinations
+    current_destinations = connectivity_statement.destinations.all()
+    new_destinations = statement[DESTINATIONS]
+
+    if len(current_destinations) != len(new_destinations):
+        return True
+
+    # We may need to change this algorithm when multi-destination is supported by neurondm
```
afonsobspinto, line 281: The issue here is that multi-destinations have no way to identify themselves, opposed to vias that have the order.
Currently I check if the set of all destinations from neurondm matches with the set of all destinations from composer (and because there's only 1 in neurondm, it will work), but if at some point there's more than one, we may be interested in a different approach. cc @ddelpiano

backend/composer/services/cs_ingestion/neurondm_script.py
zsinnema: @afonsobspinto I'm not aware that we use this so we can remove it

backend/composer/services/cs_ingestion/cs_ingestion_services.py
zsinnema: @afonsobspinto as said during the last meeting I would split the ingestion into steps. For example 
step 1 is load and transform all KS from neurondm into "an internal format" in memory
step 2 load all KS (and set the correct state of these KS) to the database
step 3 post ingestion step to update KS into (in)validate depending on depending KS  validations

I suggest that step 3 has one note per ingestion cyclus per KS with the reason that it's invalidated because of "the list of all invalid dependencies (path)"

cc: @ddelpiano

afonsobspinto: The current implementation is already following that flow. The question I have here is, in step 2 we will have connectivity statements set to the Invalid state and we will add a note to it saying why.
Then, on step 3, which is optional and configurable by a flag `update_upstream`, it can happen that one of the connectivity statements we set to Invalid in 2., also should become Invalid because of dependencies. In this case, do we add a new note?

zsinnema: imo we should add another note yes

backend/composer/services/cs_ingestion/helpers.py
```diff
@@ -0,0 +1,53 @@
...
+    else:
+        return None
+
+
+def found_entity(uri: str) -> bool:
+    return AnatomicalEntity.objects.filter(ontology_uri=uri).exists()
+
+
+def update_model_instance(instance, updates: Dict):
+    for field, value in updates.items():
```
zsinnema, line 51: @afonsobspinto question: did you consider using the `update_or_create(...,defaults={the dict})` instead of copying the attributes

zsinnema (approved):

PR #206, opened 2024-01-09
Feature/sckan 238

Salam-Dalloul:
- fix [#228](https://metacell.atlassian.net/browse/SCKAN-228)

https://github.com/MetaCell/sckan-composer/assets/32265731/0969dadd-4dae-4f1b-a5b8-dadf620c0faa

- fix [#234](https://metacell.atlassian.net/browse/SCKAN-228)

Screenshot 2024-01-09 at 2 46 00 PM

-fix [#238](https://metacell.atlassian.net/browse/SCKAN-238)

https://github.com/MetaCell/sckan-composer/assets/32265731/26566ed1-9458-48fd-81a5-7524bbc8cb16

### DIFF
--- a/frontend/src/components/Widgets/CustomEntitiesDropdown.tsx
+++ b/frontend/src/components/Widgets/CustomEntitiesDropdown.tsx
@@ -256,34 +256,41 @@ export default function CustomEntitiesDropdown({
     {},
   );
 
-  const handleSelectAll = (group: string) => {
+  const handleSelectDeselectGroup = (group: string) => {
     const newSelectedOptions = [...selectedOptions];
-    autocompleteOptions
-      .filter((option: Option) => option.group === group)
-      .forEach((item) => {
-        if (
-          !newSelectedOptions.some(
-            (selectedItem) => selectedItem.id === item.id,
-          )
-        ) {
-          newSelectedOptions.push(item);
+    const groupOptions = autocompleteOptions.filter((option: Option) => option.group === group);
+    
+    // Check if all options in this group are already selected
+    const allSelectedInGroup = groupOptions.every(
+      (groupOption) => newSelectedOptions.some((selectedOption) => selectedOption.id === groupOption.id)
+    );
+    
+    if (allSelectedInGroup) {
+      // Deselect all options in this group
+      groupOptions.forEach((option) => {
+        const index = newSelectedOptions.findIndex((selected) => selected.id === option.id);
+        if (index !== -1) {
+          newSelectedOptions.splice(index, 1);
         }
       });
+    } else {
+      // Select all options in this group
+      groupOptions.forEach((option) => {
+        if (!newSelectedOptions.some((selected) => selected.id === option.id)) {
+          newSelectedOptions.push(option);
+        }
+      });
+    }
+    
     handleSelectedOptionsChange(newSelectedOptions);
   };
-
-  const handleDeselectAll = (group: string) => {
-    const newSelectedOptions = selectedOptions.filter(
-      (item) =>
-        !autocompleteOptions
-          .filter((option: Option) => option.group === group)
-          .some((selectedItem) => selectedItem.id === item.id),
-    );
-    handleSelectedOptionsChange(newSelectedOptions);
-  };
-
+  
   const getGroupButton = (group: string) => {
-    const allObjectsExist = selectedOptions.length === allOptions.length;
+    const groupOptions = autocompleteOptions.filter((option: Option) => option.group === group);
+    const allSelectedInGroup = groupOptions.every(
+      (groupOption) => selectedOptions.some((selectedOption) => selectedOption.id === groupOption.id)
+    );
+    
     return (
       <Button
         variant="text"
@@ -293,11 +300,9 @@ export default function CustomEntitiesDropdown({
           fontWeight: 600,
           lineHeight: "1.125rem",
         }}
-        onClick={() =>
-          allObjectsExist ? handleDeselectAll(group) : handleSelectAll(group)
-        }
+        onClick={() => handleSelectDeselectGroup(group)}
       >
-        {allObjectsExist ? `Deselect` : `Select`} All
+        {allSelectedInGroup ? `Deselect all` : `Select all`}
       </Button>
     );
   };
--- a/frontend/src/components/Widgets/CustomEntitiesDropdown.tsx
+++ b/frontend/src/components/Widgets/CustomEntitiesDropdown.tsx
@@ -74,8 +74,6 @@ const styles = {
       right: "0.0625rem",
       top: "0.0625rem",
       pointerEvents: "none",
-      background:
-        "linear-gradient(270deg, #FFF 67.69%, rgba(255, 255, 255, 0.00) 116.94%)",
       borderRadius: "0 0.5rem 0.5rem 0",
     },
   },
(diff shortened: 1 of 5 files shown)


frontend/src/components/Widgets/CustomEntitiesDropdown.tsx
```diff
@@ -721,7 +744,9 @@ export default function CustomEntitiesDropdown({
                                   disableRipple
                                   icon={<UncheckedItemIcon fontSize="small" />}
                                   checkedIcon={
-                                    <CheckedItemIcon fontSize="small" />
+                                    (areArraysOfObjectsEqual(selectedOptions, preLevelItems))
```
afonsobspinto, line 747: Might be better to check if the arrays are equal just once and store that value in a variable, instead of doing that comparation every time you need

frontend/src/components/Widgets/CustomEntitiesDropdown.tsx
```diff
@@ -213,6 +211,7 @@ export default function CustomEntitiesDropdown({
     refreshStatement,
     statement,
     fieldName = "",
+    getPreLevelSelectedValues
   },
 }: any) {
```
afonsobspinto, line 216: I don't dislike it, but according to the [issue specification](https://metacell.atlassian.net/browse/SCKAN-238) there should be no badge/counter when we are connecting to all the above.

>then the from input text should be empty (no chips and no counter).

image

frontend/src/helpers/dropdownMappers.ts
```diff
@@ -121,6 +121,32 @@ export const processFromEntitiesData = (allOptions: tempOption[]) => {
 export function getViasGroupLabel(currentIndex: number | null) {
   return currentIndex ? `${ViasGroupLabel}-${currentIndex}` : OriginsGroupLabel;
 }
+export function areArraysOfObjectsEqual(a: any[], b: any[]): boolean {
```
afonsobspinto, line 124: I might be missing something, but can't we just get the ids of each object in both arrays, make a set of them, and compare if them?

frontend/src/components/Widgets/CustomEntitiesDropdown.tsx
afonsobspinto: @ddelpiano I don't dislike the no removal of the badges because I think it makes it a bit more explicitly that the from entities are not empty. What do you think?

zsinnema: @afonsobspinto imo we should remove the badges because this is also another / extra indicator that the from is empty meaning linking to all above

frontend/src/helpers/dropdownMappers.ts
zsinnema: @afonsobspinto @ddelpiano I'm not so sure if in the second video (Screen.Recording.2024-01-09.at.2.48.06.PM.mov) from time 14 seconds is correct.
as I understand from the video the initial is that the 2nd via is connected (explicit) to all on the first via and has an explicit link to 1 origin. Then the link to the origin is removed.

|What I then notice is that the links to the first vias become implicit (3rd state of the checkbox). Imo this is wrong, they should still show an explicit link (checked state)

zsinnema (changes requested):
@afonsobspinto @ddelpiano I'm not so sure if in the second video (Screen.Recording.2024-01-09.at.2.48.06.PM.mov) from time 14 seconds is correct.
as I understand from the video the initial is that the 2nd via is explicit linked to all on the first via and has an explicit link to 1 origin. Then the link to the origin is removed.

What I then notice is that the links to the first vias become implicit (3rd state of the checkbox). Imo this is wrong, they should still show an explicit link (checked state)

PS: with explicit I mean linked by a checkbox in checked state (and stored in the database)

frontend/src/helpers/dropdownMappers.ts
```diff
@@ -121,6 +121,16 @@ export const processFromEntitiesData = (allOptions: tempOption[]) => {
...
+  if (a.length !== b.length) return false;
+  
+  const setA = new Set(a.map(obj => obj.id));
+  const setB = new Set(b.map(obj => obj.id));
+  
+  return JSON.stringify([...setA].sort()) === JSON.stringify([...setB].sort());
```
afonsobspinto, line 130: I think we can avoid using JSON.stringify for comparation and instead use something like:
```
export function areArraysOfObjectsEqual(a: any[], b: any[]): boolean {
  if (a.length !== b.length) return false;

  const setA = new Set(a.map(obj => obj.id));
  const setB = new Set(b.map(obj => obj.id));

  if (setA.size !== setB.size) return false;

  for (let id of setA) {
    if (!setB.has(id)) return false;
  }

  return true;
}
```

The idea behind creating a Set is so that we can perform a check if an element is present in constant time O(1), without the need to sort (which would take more time, O(n log n))

afonsobspinto (changes requested):
https://github.com/MetaCell/sckan-composer/pull/206/files#r1447504327

frontend/src/helpers/dropdownMappers.ts
zsinnema: @afonsobspinto what is there is an id in setB that is not in setA

afonsobspinto: I think you might have missed this line @zsinnema :
```
  if (setA.size !== setB.size) return false;
```

If the sets have different sizes for sure they are not equal.
If the sets have equal size, it doesn't matter if all of B are in A or if all of A are in B.

zsinnema: @afonsobspinto agree :-)

frontend/src/helpers/dropdownMappers.ts
```diff
@@ -127,7 +127,13 @@ export function areArraysOfObjectsEqual(a: any[], b: any[]): boolean {
   const setA = new Set(a.map(obj => obj.id));
   const setB = new Set(b.map(obj => obj.id));
   
-  return JSON.stringify([...setA].sort()) === JSON.stringify([...setB].sort());
+  if (setA.size !== setB.size) return false;
```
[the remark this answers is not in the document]
zsinnema: @Salam-Dalloul the algo of @afonsobspinto works. so all good here :-)

ddelpiano:
> #238

@zsinnema this is behaving according to requirements, unless the client ask us to change this again to me the behaviour is what we have been asked to implement. Thanks

cc: @Salam-Dalloul

zsinnema:
@ddelpiano I think you didn't get my comment or I don't know the exact history of the CS via history

if the user explicit links a via to all the objects one level above IS NOT THE SAME as not linking the via
the difference is that if a user adds a new element in the level above this is NOT automatically linked to the via (via a virtual link)
this is exactly represented in the 3 state checkbox: checked --> explicit linked, unchecked --> not linked, "box" --> implicit linked

The already uploaded video imo shows the explicit linked via's of the first level (explicit because they have the checked state) are converted to implicit linked via's (the 3rd checkbox state)

It could be that the video is not "correct" in a way that the history of the via before it was linked to the origin (this action makes the whole show as checked checkboxes (meaning explicit) is not shown).

@Salam-Dalloul could you upload a video with these 2 scenarios:
1. the second level via is not linked to the first level + then the via is linked to the origin + then unlink the origin
2. the second level via is explicit linked to all via's in the first level + then the via is linked to the origin + then unlink the origin

thanks!

ddelpiano:
@Salam-Dalloul I would expect that when I click on a connection they use the checked state since these are becoming explicit, check the video out.

[scrnli_11_01_2024_18-01-29.webm](https://github.com/MetaCell/sckan-composer/assets/37704177/20d1af9b-5eed-40a4-91e2-85826cb99fd4)

Salam-Dalloul:
> @Salam-Dalloul I would expect that when I click on a connection they use the checked state since these are becoming explicit, check the video out.
> 
>  scrnli_11_01_2024_18-01-29.webm

@ddelpiano fixed

ddelpiano (approved):

zsinnema (approved):
lgtm

PR #211, opened 2024-01-25
Feature/sckan 252

afonsobspinto:
Closes https://metacell.atlassian.net/browse/SCKAN-252

### DIFF
--- a/backend/composer/services/graph_service.py
+++ b/backend/composer/services/graph_service.py
@@ -4,7 +4,7 @@ JOURNEY_DELIMITER = '\\'
 def generate_paths(origins, vias, destinations):
     paths = []
     # Calculate the total number of layers, including origins and destinations
-    number_of_layers = len(vias) + 2 if vias else 2
+    destination_layer = max([via.order for via in vias] + [0]) + 2 if vias else 1
 
     # Handle direct connections from origins to destinations
     for origin in origins:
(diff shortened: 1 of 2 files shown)


backend/composer/services/graph_service.py
```diff
@@ -42,8 +42,8 @@ def create_paths_from_origin(origin, vias, destinations, current_path, number_of
         # This checks if the last node in the current path is one of the nodes that can lead to the current via.
         # In other words, it checks if there is a valid connection
         # from the last node in the current path to the current via.
-        if len(current_path) == via_layer and (current_path[-1][0] in list(
-            a.name for a in current_via.from_entities.all()) or not current_via.from_entities.exists()):
+        if current_path[-1][0] in list(
```
zsinnema, line 45: question: @afonsobspinto i've added this check becaause the diagram was showing wrongly when the sort of the via is not consecutive and / or doesn't start with 0. Can you test  this?

afonsobspinto: It should be impossible, unless we do direct database operations, to create non consecutive and / or doesn't start with 0 due to our [custom save](https://github.com/MetaCell/sckan-composer/blob/develop/backend/composer/models.py#L617-L629).

With that said, I added a simple test case to check that extraordinary case and I found a problem with the destination layer which I fixed. (I'm not sure if that's the type of problems you were expecting or if I need to have a more complex example to trigger it)

zsinnema: @afonsobspinto can you write down the test data of your test? thanks

afonsobspinto: [This](https://github.com/MetaCell/sckan-composer/blob/fc3b4fb1bf1d774b13934963c19761d8aee6537c/backend/tests/test_journey.py#L457) is the unit test I was referring to.

Here's the 'graph' representation:
image

PR #214, opened 2024-01-30
#230 display note time depends on the user local timezone

Salam-Dalloul:


frontend/src/helpers/helpers.ts
```diff
@@ -153,30 +154,38 @@ export const formatTime = (date: string) => {
...
-  const minutes = Math.floor(seconds / 60);
-  const hours = Math.floor(minutes / 60);
-  const days = Math.floor(hours / 24);
-  const weeks = Math.floor(days / 7);
-  const years = Math.floor(weeks / 52);
-
+  const machineTimeZone = moment.tz.guess();
+  
+  const timestampMoment = moment.tz(timestamp, 'CET');
+  const now = moment.tz(machineTimeZone);
```
zsinnema, line 160: @Salam-Dalloul imho we don't need an external lib

what do you think of this:
```
const cetTimeZoneOffset = 3600000;
timestampInCurrentTZ = new Date(new Date(timestamp).getTime() + cetTimeZoneOffset + (new Date().getTimezoneOffset() * 60000))
```

zsinnema (changes requested):

frontend/src/helpers/helpers.ts
zsinnema: can you explain what this statement does

Salam-Dalloul: @zsinnema thanks for the suggestion, as i understood, it converts the timestamp into the user timezone by first adds the 1 hour different between the UTC and CET tz by adding this diff 1 hour to the timestamp date object  `new Date(timestamp).getTime()` and then it adds the offset of the current tz to adjust the timestamp to the local tz!

zsinnema: @Salam-Dalloul yes nice that's correct :-) and i believe that's exactly what we need here and therefore we don't need this external package

zsinnema (approved):

PR #219, opened 2024-02-09
SCKAN-262: fix UI crash for button menu

ddelpiano:


Salam-Dalloul (dismissed):

frontend/src/components/Widgets/CustomGroupedButtons.tsx
```diff
@@ -40,6 +40,7 @@ export default function GroupedButtons({handleClick, selectedOption, options, se
           aria-label="select merge strategy"
           aria-haspopup="menu"
           onClick={handleToggle}
+          disabled={options?.length > 0 ? false : true}
```
Salam-Dalloul, line 43: hey @ddelpiano i only have one suggestion here to make the button looks better, for now it looks like this
Screenshot 2024-02-09 at 2 54 16 PM

if we just make an edit for the options passed value to have an initial value as an empty array if it's undefined in here 
`export default function GroupedButtons({handleClick, selectedOption, options = [], selectedIndex, handleMenuItemClick , hasFormat = false, format, disabled}: any) {
`

and then add the disabled state for both ButtonGroup and the Arrow button like this 

` <ButtonGroup variant="contained" ref={anchorRef} disabled={options?.length <= 0 || disabled}>
        <Button onClick={handleClick}>{selectedOption}</Button>
        <Button
          size="small"
          aria-controls={open ? 'split-button-menu' : undefined}
          aria-expanded={open ? 'true' : undefined}
          aria-label="select merge strategy"
          aria-haspopup="menu"
          onClick={handleToggle}
          disabled={options?.length <= 0}
        >
          <ArrowDropDownIcon />
        </Button>
      </ButtonGroup>`

then the button when looks like this 
Screenshot 2024-02-09 at 2 55 00 PM

zsinnema (approved):
lgtm

PR #222, opened 2024-02-09
#261 remove duplicates from forward connections

Salam-Dalloul:


ddelpiano (changes requested):
lint issues to be fixed

PR #249, opened 2024-03-14
Feature/sckan 274

afonsobspinto:
closes https://metacell.atlassian.net/browse/SCKAN-274

- Adds synonym model
- Adds migration that creates synonym model
- Adds migration that moves duplicated anatomical entities to the synonyms. Keeps as primary anatomical entity name, the first name found. 
- Adds migration to make ontology uri unique for anatomical entities
- Adds migration to remove unique name constraint for anatomical entities (after discussion with @ddelpiano)
- Updates django ingest anatomical entities command to work with the new file and with the new model. Adds update_names flag to allow overwrite of the primary anatomical entities label.

See https://metacell.atlassian.net/browse/SCKAN-274?focusedCommentId=14056 for some tests


backend/composer/management/commands/ingest_anatomical_entities.py
```diff
@@ -1,64 +1,88 @@
...
+BULK_LIMIT = 100
 
 
 class Command(BaseCommand):
-    help = "Ingests Anatomical Entities CSV file(s)"
+    help = "Ingests Anatomical Entities CSV file(s), with robust error handling."
 
     def add_arguments(self, parser):
         parser.add_argument("csv_files", nargs="+", type=str)
+        parser.add_argument("--update_names", action='store_true',
```
zsinnema, line 19: question: @afonsobspinto what is the use case for this option? I don't see any benefit of setting this to False

backend/composer/management/commands/ingest_anatomical_entities.py
```diff
@@ -1,64 +1,88 @@
...
+                            help="Update the name even if the entity already exists")
+        parser.add_argument("--show_complete_logs", action='store_true',
+                            help="Show detailed logs during processing")
+
+    def _process_anatomical_entity(self, name, ontology_uri, synonym, update_names, show_complete_logs, processed_uris,
+                                   synonym_accumulator):
+        try:
+            is_first_occurrence = ontology_uri not in processed_uris
```
zsinnema, line 27: question: @afonsobspinto what is the use case for the is first occurence? What if the next occurence of the (same) uri has a different name? I don't see any error handling for that and I also don't see any logging. Imo this is either over engineering or non finiished functionality

backend/composer/management/commands/ingest_anatomical_entities.py
```diff
@@ -1,64 +1,88 @@
...
-        #     anatomical_entity.save()
+            if not created and update_names and is_first_occurrence:
+                if anatomical_entity.name != name:
+                    anatomical_entity.name = name
+                    anatomical_entity.save()
+                    if show_complete_logs:
+                        self.stdout.write(
+                            self.style.SUCCESS(f"Updated {anatomical_entity.ontology_uri} name to {name}."))
 
+            processed_uris.add(ontology_uri)
```
zsinnema, line 41: remark: @afonsobspinto imo we don't need to track processed uris

backend/composer/models.py
```diff
@@ -229,17 +229,19 @@ class AnatomicalEntity(models.Model):
...
         ordering = ["name"]
         verbose_name_plural = "Anatomical Entities"
-        constraints = [
-            models.UniqueConstraint(Upper("name"), name="ae_unique_upper_name")
-        ]
+
+
+class Synonym(models.Model):
+    anatomical_entity = models.ForeignKey(AnatomicalEntity, on_delete=models.CASCADE, related_name="synonyms")
+    alias = models.CharField(max_length=200, db_index=True)
```
zsinnema, line 244: suggestion: @afonsobspinto rename alias to name to be more in line of other models. The model name Synonym already explains what the data is

zsinnema: critical: @afonsobspinto imo the "name" of the synonym should be unique. At least within the AE.
question: @afonsobspinto what if the name of the synonym also exists in another AE?

backend/composer/management/commands/ingest_anatomical_entities.py
afonsobspinto: 👍

backend/composer/management/commands/ingest_anatomical_entities.py
afonsobspinto: The name we want to keep is the **first** one that appears. 

If the next occurrence of the same uri has a different name we do **not** want to update the name.

See [this thread ](https://metacell.slack.com/archives/C0477SPSQQK/p1710368649332379?thread_ts=1710352121.100009&cid=C0477SPSQQK)

backend/composer/management/commands/ingest_anatomical_entities.py
afonsobspinto: See https://github.com/MetaCell/sckan-composer/pull/249#discussion_r1528814178 for the explanation on why I believe we need it

backend/composer/models.py
afonsobspinto: Agree with the renaming and with the unique together. ✅ 

For the question, the data suggests that having the same name/synonym for different uris is something that can happen (see example below). 
I would say the onus of dealing with it properly is on the curator but we can ask the client cc @ddelpiano .
```
http://purl.obolibrary.org/obo/UBERON_0000019,camera-type eye,eye
http://purl.obolibrary.org/obo/UBERON_0010230,eyeball of camera-type eye,eye
```

zsinnema (approved):

PR #252, opened 2024-03-29
Feature/csckan-277

D-GopalKrishna:
Jira Link - https://metacell.atlassian.net/browse/SCKAN-277

Implementation/Requirements :
- show AE name instead of synonyms in Connected From 
- show (region, layer) uri in identifier when it is RegionLayer AE

When the destination is one of RegionLayer AE, it shows as -> Region, Layer in the "Connected from". This is since RegionLayerPair is also AnatomicalEntity.
image

@ddelpiano thoughts?

When origin is a RegionLayer AE, the identifier is a pair separated by a comma. uri for - region, layer
Screenshot 2024-03-29 122140

### DIFF
--- a/backend/composer/models.py
+++ b/backend/composer/models.py
@@ -269,6 +269,18 @@ class AnatomicalEntity(models.Model):
             return f"{region_name},{layer_name}"
         else:
             return "Unnamed Entity"
+        
+    @property
+    def ontology_uri(self):
+        if self.simple_entity:
+            return self.simple_entity.ontology_uri
+        elif self.region_layer:
+            layer_uri = self.region_layer.layer.ontology_uri if self.region_layer.layer else ""
+            region_uri = self.region_layer.region.ontology_uri if self.region_layer.region else ""
+            return f"{region_uri},{layer_uri}"
+        else:
+            return None
+
 
     def __str__(self):
         return self.name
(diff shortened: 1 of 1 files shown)


backend/composer/models.py
```diff
@@ -269,6 +269,18 @@ def name(self):
             return f"{region_name},{layer_name}"
         else:
             return "Unnamed Entity"
+        
+    @property
+    def ontology_uri(self):
+        if self.simple_entity:
```
zsinnema, line 275: question: @D-GopalKrishna can you think of a solution that is a oneline solution?
please check with @afonsobspinto, I asked him yesterday the same :-)

zsinnema (changes requested):

D-GopalKrishna:
Simplification of the code in the image - valid, since we have a code for constraint already in place as following

```
    class Meta:
        constraints = [
            CheckConstraint(
                check=(
                        Q(simple_entity__isnull=False, region_layer__isnull=True) |
                        Q(simple_entity__isnull=True, region_layer__isnull=False)
                ),
                name='check_anatomical_entity_exclusivity'
            )
        ]
```
by @afonsobspinto 


image


So we never have the condition for the "else". Also the region and layer have `required` conditions.

PR #254, opened 2024-04-02
Feature/csckan-276 - Modify the rest API to expose the new region+layer

D-GopalKrishna:
Jira link - https://metacell.atlassian.net/browse/SCKAN-276

Changes made:
**Backend**
- Update serializer to show Synonyms, models to have annotated_name (from [django_filters](https://django-filter.readthedocs.io/en/stable/ref/filters.html#multiplechoicefilter)), and update openapi spec
- Update filterset logic - to merge the queryset made by filters of name, URI, and synonyms. 

**Frontend**
- Show Layer Region URI and label based on logic if region_layer exists. Else show simple entity with just ontology_uri
- changes in the statement form footer for region layer URI

Short video demo

https://github.com/MetaCell/sckan-composer/assets/59233227/5cd77357-1fa0-4e90-97e0-d1657ad4139c

Statement form Footer for region layer URI
image


D-GopalKrishna:
@ddelpiano - Contrast from the Jira Image shown. 
image

D-GopalKrishna:
LOG: Previous implementation for the Q() chaining 

image

Replaced with the new - UNION of queryset. 
- Generally UNION is more efficient than OR. It can use indexes much better for each queryset and later unionize the results.

backend/composer/api/serializers.py
```diff
@@ -673,3 +679,19 @@ class Meta:
             "statement_preview",
             "errors"
         )
+
+
+class GenericConnectivityStatementSerializer(ConnectivityStatementSerializer):
```
zsinnema, line 684: @D-GopalKrishna see also my previous comment about the name `Generic`

backend/composer/models.py
```diff
@@ -120,6 +123,24 @@ def get_queryset(self):
         )
 
 
+class AnnotatedConnectionLayerQuerySet(models.QuerySet):
+    """
+    Useful for filtering - GenericConnectionStatementFilter
+    """
+    def annotate_with_ontology_uri(self):
```
zsinnema, line 130: remark: @D-GopalKrishna it is nice to use `annotate` to create a new result (column) in the query

critical: @D-GopalKrishna however I also see you using the annotated result in a filter. This is not a good idea because it breaks the indexing.... unless you create a functional index in the postgresql database.. As expllained last week please construct a union for this: query filter by `anatomical_entities__simple_entity__ontology_uri` union filter by `anatomical_entities__region_layer__region__ontology_uri` union filter by `anatomical_entities__region_layer__layer__ontology_uri`

backend/composer/models.py
```diff
@@ -120,6 +123,24 @@ def get_queryset(self):
...
+        return self.annotate(annotated_ontology_uri=models.Case(
+            models.When(anatomical_entities__simple_entity__isnull=False, then=models.F('anatomical_entities__simple_entity__ontology_uri')),
```
zsinnema, line 132: @D-GopalKrishna same here

backend/composer/models.py
```diff
@@ -254,11 +275,46 @@ class Meta:
...
+    def __str__(self):
+        return f"{self.region.name} - {self.layer.name}"
+
+
+class AnnotatedAnatomicalEntityQuerySet(models.QuerySet):
+    """
+    Useful for filtering - GenericConnectionStatementFilter
+    """
+    def annotate_with_ontology_uri(self):
+        return self.annotate(annotated_ontology_uri=models.Case(
```
zsinnema, line 287: @D-GopalKrishna and here

backend/composer/models.py
```diff
@@ -254,11 +275,46 @@ class Meta:
...
+                models.F('region_layer__region__ontology_uri'),
+                models.Value(','),
+                models.F('region_layer__layer__ontology_uri')
+            )),
+            default=models.Value(''),
+            output_field=models.URLField()
+        ))
+
+    def annotate_with_name(self):
+        return self.annotate(annotated_name=models.Case(
```
zsinnema, line 299: @D-GopalKrishna :-) and here

backend/composer/models.py
```diff
@@ -269,6 +325,18 @@ def name(self):
...
+    @property
+    def ontology_uri(self):
+        if self.simple_entity:
+            return self.simple_entity.ontology_uri
+        elif self.region_layer:
+            layer_uri = self.region_layer.layer.ontology_uri if self.region_layer.layer else ""
+            region_uri = self.region_layer.region.ontology_uri if self.region_layer.region else ""
+            return f"{region_uri},{layer_uri}"
+        else:
+            return None
```
zsinnema: comment: @D-GopalKrishna I think this else will never be run (it's either simple or region layer)

backend/composer/models.py
```diff
@@ -269,6 +325,18 @@ def name(self):
             return f"{region_name},{layer_name}"
         else:
             return "Unnamed Entity"
+        
+    @property
+    def ontology_uri(self):
+        if self.simple_entity:
+            return self.simple_entity.ontology_uri
```
zsinnema: suggestion: @D-GopalKrishna why not do something Pythonic like this: `return simple if simple else region layer`

zsinnema (changes requested):

backend/composer/models.py
D-GopalKrishna: Yes indeed, already did - https://github.com/MetaCell/sckan-composer/pull/252#issuecomment-2031215692. And merged to this branch as well.

backend/composer/models.py
D-GopalKrishna: Yes indeed, already did - https://github.com/MetaCell/sckan-composer/pull/252#issuecomment-2031215692. And merged to this branch as well.

backend/composer/api/views.py
```diff
@@ -345,6 +347,32 @@ def partial_update(self, request, *args, **kwargs):
         return super().partial_update(request, *args, **kwargs)
 
 
+class GenericConnectivityStatementViewSet(
```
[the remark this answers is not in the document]
D-GopalKrishna: Generic wil be a new endpoint that is:
- public
- only allows GET request
- has different filterset
- returns a different kind of response - hence a different serializer.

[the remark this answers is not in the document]
zsinnema: @D-GopalKrishna imo we should "group" (OpenAPI Tags) these kind of endpoints 
like this (add to the view)
```
@extend_schema(
    tags=["public",]
)
```
and imo the ConnectivityStatements are the "published" ones so I would rather go for a naming that matches close to the usage instead on the `Generic` name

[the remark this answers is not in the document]
D-GopalKrishna: Done! For keeping it consistent with the state- does ExportedConnectivityStatement sound good?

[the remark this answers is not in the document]
D-GopalKrishna: Will be addressing this on #35 PR.

zsinnema (approved):

PR #260, opened 2024-04-10
Feature/esckan-35 - Composer REST new generic endpoint - public - KnowledgeStatement

D-GopalKrishna:
Jira Link - https://metacell.atlassian.net/browse/ESCKAN-35

A new generic endpoint for public Knowledge statement.

Add new endpoint with new serializer - KnowledgeStatementSerializer and filterset - KnowledgeStatementFilterSet.

Only GET is allowed - is tagged "public".
image

filter by 4 query parameters - origin, via, destination, population uris
Only exported Connectivity statements are shown. As mentioned in [lucidcharts](https://lucid.app/lucidchart/7197f056-4f35-4453-8f27-369a045ad40b/edit?invitationId=inv_d553a755-dc0e-44bc-ac75-f15e4be38e13&page=.eqhIXv27vP.#)

Default behaviors of the filters. Example:
- When multiple population uris are selected - OR is applied.
- When one population uri and one via/origin/destination is selected - AND is applied


backend/composer/admin.py
```diff
@@ -114,6 +114,11 @@ def get_queryset(self, request: HttpRequest) -> QuerySet[Any]:
     def synonyms(self, obj):
         synonyms = obj.synonyms.all()
         return ', '.join([synonym.name for synonym in synonyms])
+
```
D-GopalKrishna, line 117: Question from @zsinnema : - 

> question: @D-GopalKrishna is there a special reason why you didn't add it to the DefaultRouter (created on line 20)?


Yes correct! Fixed

zsinnema (approved):

backend/composer/api/serializers.py
```diff
@@ -680,3 +680,18 @@ class Meta:
             "errors"
         )
 
+
+class KnowledgeStatementSerializer(ConnectivityStatementSerializer):
+    """Knowledge Statement"""
+    class Meta(ConnectivityStatementSerializer.Meta):
+        fields = (
```
afonsobspinto, line 687: suggestion: I think we should add the reference_uri here since that's one of the fields we allow filtering on.

PR #261, opened 2024-04-10
frontend styling and some journey improvements

ddelpiano:
SCKAN-278 and SCKAN-283

### DIFF
--- a/backend/composer/services/graph_service.py
+++ b/backend/composer/services/graph_service.py
@@ -4,6 +4,13 @@ from django.apps import apps
 JOURNEY_DELIMITER = '\\'
 
 
+def get_entity_name(entity):
+    if entity.region_layer is not None:
+        return f"{entity.region_layer.region.name} ({entity.region_layer.layer.name})"
+    else:
+        return entity.name
+
+
 def generate_paths(origins, vias, destinations):
     paths = []
     # Calculate the total number of layers, including origins and destinations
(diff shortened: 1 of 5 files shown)


backend/composer/services/graph_service.py
```diff
@@ -4,6 +4,13 @@
 JOURNEY_DELIMITER = '\\'
 
 
+def get_entity_name(entity):
```
afonsobspinto, line 7: suggestion: This is most likely introduced before https://github.com/MetaCell/sckan-composer/pull/257/commits/01bba38fc65ee6017c14b0dd6aa1ebbb9b46d61c got merged into develop. With the name being a @property of AnatomicalEntity I don't expect us to need this `get_entity_name` anymore

zsinnema: @afonsobspinto I agree

ddelpiano: I don't agree, the above makes the naming we use in the dropdown consistent with the journey and makes the journey/preview more readable.

Current preview:
```
A connection goes from descending colon,Myenteric nerve plexus of descending colon to inferior mesenteric ganglion via descending colon,Longitudinal muscle layer of descending colon via descending colon,Serosa of descending colon via lumbar colonic nerve.
```

with get_entity_name:
```
A connection goes from descending colon (Myenteric nerve plexus of descending colon) to inferior mesenteric ganglion via descending colon (Longitudinal muscle layer of descending colon) via descending colon (Serosa of descending colon) via lumbar colonic nerve.
```

ddelpiano: @afonsobspinto I introduced that method yesterday since the current name for AE is region,layer and it makes difficult read/interpret the journey since that is the same separator we use for multiple entities in a layer, I get lost of what is a simple entity, what a region and what a layer.
With that method I am just respecting the convention we put in the acceptance criteria to express region (layer) through the UI, but since this is generated backend wise that is why I added such method.

zsinnema: @ddelpiano but shouldn't we always use the same convention? Is so then i suggest to fix/change the commit if @afonsobspinto

ddelpiano: @zsinnema fine by me

afonsobspinto: I just changed the name property to match with @ddelpiano request:

image

afonsobspinto (approved):

backend/composer/services/graph_service.py
ddelpiano: lgtm

zsinnema (approved):
Lgtm

PR #264, opened 2024-04-11
Re-aligning graph styling with sckan explorer

ddelpiano:


frontend/src/components/ProofingTab/GraphDiagram/Factories/CustomNodeFactory.tsx
```diff
@@ -25,9 +25,9 @@ export class CustomNodeFactory extends AbstractReactFactory {
             case NodeTypes.Via:
                 return <ViaNodeWidget engine={this.engine} model={event.model} />;
             case NodeTypes.Destination:
-                return <DestinationNodeWidget engine={this.engine} model={event.model} />;
+                return <DestinationNodeWidget engine={this.engine} model={event.model} forwardConnection={false}/>;
```
afonsobspinto, line 28: issue (blocking): My understanding is that forwardConnection shouldn't always be false but it's rather on event.model.options.forward_connection.length

afonsobspinto (changes requested):
Looks very nice, just added a question/issue

frontend/src/components/ProofingTab/GraphDiagram/Factories/CustomNodeFactory.tsx
afonsobspinto: Potentially we don't even need the prop as we already have access to the model which contains the forward_connection in the options but that's a detail. The issue I'm raising is the hardcoded false value

afonsobspinto (changes requested):
issue (blocking): DestinationNodeWidget needs to be adapted to no longer rely on the forwardConnection prop but instead on model.options.forward_connection.

afonsobspinto (approved):

PR #265, opened 2024-04-17
Feature/esckan-35 -  KnowledgeStatement endpoint fixes

D-GopalKrishna:
jira Link - update on -> https://metacell.atlassian.net/browse/ESCKAN-35

- add reference_uri to KnowledgeStatement endpoint response 
- Corrected the serializer to_representation for KS-serializer - `super(ConnectivityStatementSerializer, self).to_representation(instance)`
- Added CORS to `/api/composer/knowledge-statement/` using [django-cors-header's signals](https://github.com/adamchainz/django-cors-headers?tab=readme-ov-file#signals).

### DIFF
--- a/backend/composer/api/serializers.py
+++ b/backend/composer/api/serializers.py
@@ -683,6 +683,18 @@ class ConnectivityStatementUpdateSerializer(ConnectivityStatementSerializer):
 
 class KnowledgeStatementSerializer(ConnectivityStatementSerializer):
     """Knowledge Statement"""
+    def to_representation(self, instance):
+        representation = super(ConnectivityStatementSerializer, self).to_representation(instance)
+        depth = self.context.get('depth', 0)
+
+        if depth < 1:
+            representation["forward_connection"] = KnowledgeStatementSerializer(
+                instance.forward_connection.all(),
+                many=True,
+                context={**self.context, 'depth': depth + 1}
+            ).data
+        return representation
+    
     class Meta(ConnectivityStatementSerializer.Meta):
         fields = (
             "id",
@@ -696,3 +708,4 @@ class KnowledgeStatementSerializer(ConnectivityStatementSerializer):
             "phenotype",
             "reference_uri"
         )
+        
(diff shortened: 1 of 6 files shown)


backend/composer/api/serializers.py
```diff
@@ -683,6 +683,11 @@ class Meta:
 
 class KnowledgeStatementSerializer(ConnectivityStatementSerializer):
     """Knowledge Statement"""
+    def to_representation(self, instance):
```
zsinnema, line 686: suggestion: @D-GopalKrishna can't we add ```
exclude = ('forward_connection', )
``` to the inner Meta class?

ddelpiano (changes requested):
forward connections it's one of the information we need, please re-include it. Thanks :)

backend/composer/api/serializers.py
D-GopalKrishna: thanks @zsinnema! made change as @ddelpiano mentioned.

afonsobspinto:
suggestion: Should we allow this endpoint to be accessible from any origin? In other words, can Explorer make requests to this endpoint directly or does it need a reverse-proxy?

D-GopalKrishna:
> question: Should we allow this endpoint to be accessible from any origin (enable CORS)?
> 
> Specifically, will Explorer be able to make requests directly to this endpoint or will it need a proxy?

Indeed @afonsobspinto! We do need to allow CORS.

zsinnema:
@afonsobspinto imo we don't need cors

D-GopalKrishna:
Edited the title and description of the PR to include the full scope of the fix for KS endpoint.

D-GopalKrishna:
About CORS from the docs. - https://github.com/adamchainz/django-cors-headers?tab=readme-ov-file#signals

image

<br>

In our case, we don't use `CORS_ALLOWED_ORIGINS`. We simply use `from corsheaders.signals import check_request_enabled` to enable the path for `/api/composer/knowledge-statement/`.

cc @zsinnema @afonsobspinto

zsinnema (approved):

PR #266, opened 2024-04-17
editing business logic to remove condition not anymore required

ddelpiano:


zsinnema (changes requested):
critical: @ddelpiano the removal of the sex being mandatory leads to a breaking issue in the export. The get rows will raise this exception: https://github.com/MetaCell/sckan-composer/blob/a002cd6bd87f581d8159be0059b2446a582a659c/backend/composer/services/export_services.py#L400 which will mark the ConnectivityStatement as invalid
please fix the issue

PR #286, opened 2024-06-18
Feature/esckan 58

afonsobspinto:
Closes https://metacell.atlassian.net/browse/ESCKAN-58

- Updates KnowledgeStatement serializer (and ConnectivityStatement serializer) to remove memoized 'journey' after representation is calculated.

For context the journey is being memorized per instance because it's a calculated property which is somewhat expensive to calculate and it's used both in the journey attribute itself and in the statement preview. 
The problem fixed with this change was not present in Composer because we only used the ConnectivityStatement serializer for individual serializations (not on 'list' actions).

The fix should have no impact in composer.

image

https://github.com/MetaCell/sckan-composer/assets/19196034/7f1d3593-7ef2-4458-b4e1-06956e656330

### DIFF
--- a/backend/composer/api/serializers.py
+++ b/backend/composer/api/serializers.py
@@ -597,6 +597,10 @@ class ConnectivityStatementSerializer(BaseConnectivityStatementSerializer):
                 many=True,
                 context={**self.context, 'depth': depth + 1}
             ).data
+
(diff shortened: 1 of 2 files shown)


ddelpiano (approved):
lgtm

PR #287, opened 2024-06-27
Feature/csckan-293 - Front end Projection Phenotypes do not reflect Projection phenotypes specified in Admin table

D-GopalKrishna:
Jira ticket - [SCKAN-293](https://metacell.atlassian.net/browse/SCKAN-293)

Task description:
1. Update serializer, viewset, and router - to avail ProjectionPhenotype.
2. Generate the openapi spec.
3. Replace projection phenotype instead of projection (from enum) in the FE 
4. remove projection from admin - so user can only add projection phenotype in the admin panel.

Video demo

https://github.com/MetaCell/sckan-composer/assets/59233227/01188665-82a4-45ce-a68b-ac0633e13ec7

[SCKAN-293]: https://metacell.atlassian.net/browse/SCKAN-293?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/composer/admin.py
+++ b/backend/composer/admin.py
@@ -194,6 +194,7 @@ class ConnectivityStatementAdmin(
         "sentence__pmcid",
         "knowledge_statement",
     )
+    exclude = ("projection",)
 
     fieldsets = ()
 
--- a/backend/composer/api/serializers.py
+++ b/backend/composer/api/serializers.py
@@ -11,6 +11,7 @@ from ..enums import SentenceState, CSState
 from ..models import (
     AnatomicalEntity,
     Phenotype,
+    ProjectionPhenotype,
     Sex,
     ConnectivityStatement,
     Provenance,
@@ -176,6 +177,14 @@ class PhenotypeSerializer(UniqueFieldsMixin, serializers.ModelSerializer):
 
     class Meta:
         model = Phenotype
+        fields = ("id", "name")
+
+
+class ProjectionPhenotypeSerializer(UniqueFieldsMixin, serializers.ModelSerializer):
+    """Phenotype"""
+
+    class Meta:
+        model = ProjectionPhenotype
         fields = ("id", "name")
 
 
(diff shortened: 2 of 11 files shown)


ddelpiano (changes requested):
@D-GopalKrishna great job but there is one thing missing.
I don't see the projection phenotype in the export at the moment, could you please fix that? thanks

D-GopalKrishna:
The current version exports the ProjectionPhenotype, as shown in the image below @ddelpiano 

image

backend/composer/admin.py
```diff
@@ -194,6 +194,7 @@ class ConnectivityStatementAdmin(
         "sentence__pmcid",
         "knowledge_statement",
     )
+    exclude = ("projection",)
```
afonsobspinto, line 197: question (non-blocking): Is there any reason for us to keep projection as a property of the connectivity statement? My understanding is that it was being **replaced** by ProjectionPhenotype

afonsobspinto: follow-up question: If it's meant to replace. How do we map the previous Projection values to the new ProjectionPhenotype?

ddelpiano: it seems to me that this values hasn't been used so far, looking at the dev and prod data.
Plus the ingestion does not seem is bringing this info in, so we should be fine @afonsobspinto

ddelpiano (approved):

afonsobspinto (approved):

backend/composer/admin.py
afonsobspinto: In that case I would suggest to remove the projection attribute from our model. This is a non-blocking suggestion

ddelpiano: I'll add this as tech debt for the next contract.

PR #288, opened 2024-07-09
Feature/csckan-290 - change region/layer model to be independent models from AEMeta - and related changes

D-GopalKrishna:
Jira link - [SCKAN-290](https://metacell.atlassian.net/browse/SCKAN-290)

Task description
1. Update the models 
2. do the migrations + data migrations -> migrations are divided into three steps along with useful comments - steps taken for each step.
3. admin changes with the new models
4. signal modifications to accommodate the new models. 
5. Ingestion script updated
6. validator logic updated

[SCKAN-290]: https://metacell.atlassian.net/browse/SCKAN-290?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


D-GopalKrishna:
How to create a layer - 

https://github.com/MetaCell/sckan-composer/assets/59233227/03a75707-eb64-4fc5-bae4-4981e59b9d3b

D-GopalKrishna:
We don't have on_save to create other entities from the intersection table or layer/table because - each has its UI in the admin interface and saves it when saved from the admin UI. 

Here is a video demo

https://github.com/MetaCell/sckan-composer/assets/59233227/0e29959f-96d9-41d2-8d73-084637309d01

backend/composer/models.py
```diff
@@ -240,18 +241,66 @@ class Meta:
         verbose_name_plural = "Anatomical Entities"
 
 
-class Layer(AnatomicalEntityMeta):
-    ...
+class Layer(models.Model):
+    layer_id = models.BigAutoField(primary_key=True, auto_created=True)
+    ae_meta = models.ForeignKey(AnatomicalEntityMeta, on_delete=models.CASCADE, related_name='layer_meta', null=True)
```
afonsobspinto, line 246: issue (blocking): null should be set to False. A layer should **always** be associated with an anatomical entity meta

backend/composer/models.py
```diff
@@ -240,18 +241,66 @@ class Meta:
...
+        if Layer.objects.filter(ae_meta=self.ae_meta).exists():
+            raise ValidationError('Layer with this ontology_uri already exists')
+        if Region.objects.filter(ae_meta=self.ae_meta).exists():
+            raise ValidationError('Region with this ontology_uri already exists')
+
+
+    def save(self, *args, **kwargs):
+        if Layer.objects.filter(ae_meta=self.ae_meta).exists():
+            raise ValueError('Layer with this Anatomical Entity Meta already exists')
+        elif Region.objects.filter(ae_meta=self.ae_meta).exists():
```
afonsobspinto, line 261: issue (blocking): I believe we should allow a Layer to be created even if a Region already exists for the same anatomical entity meta. We only want to avoid duplicates within each table, this means no Layers repeated, no Regions repeated and no anatomical entities meta repeated but it's fine to have an anatomical entity associated with both a region and a layer.

backend/composer/models.py
```diff
@@ -240,18 +241,66 @@ class Meta:
...
-class Layer(AnatomicalEntityMeta):
-    ...
+class Layer(models.Model):
+    layer_id = models.BigAutoField(primary_key=True, auto_created=True)
+    ae_meta = models.ForeignKey(AnatomicalEntityMeta, on_delete=models.CASCADE, related_name='layer_meta', null=True)
+
+    def __str__(self):
+        return self.ae_meta.name
+    
+    def clean(self):
```
afonsobspinto, line 251: suggestion (non-blocking): Clean and save seem to be doing the same checks. It might make sense to replace the save checks with a call to clean.

backend/composer/models.py
```diff
@@ -240,18 +241,66 @@ class Meta:
...
+
 
 
-class Region(AnatomicalEntityMeta):
-    ...
-    layers = models.ManyToManyField(Layer, through='AnatomicalEntityIntersection')
+        
+
+
+class Region(models.Model):
```
afonsobspinto, line 275: Same comments that I left for Layer can be applied to Region

backend/composer/services/cs_ingestion/helpers/anatomical_entities_helper.py
```diff
@@ -69,19 +69,19 @@ def add_entity_to_instance(instance, entity_field, entity, update_anatomic_entit
...
+        region = Region.objects.get(ae_meta__ontology_uri=region_uri)
     except Region.DoesNotExist:
         region = None
 
     if update_anatomical_entities:
         if not layer:
             try:
-                layer_meta = AnatomicalEntityMeta.objects.get(ontology_uri=layer_uri)
+                layer_meta = AnatomicalEntityMeta.objects.get(ontology_uri=layer_uri)                    
                 layer, _ = convert_anatomical_entity_to_specific_type(layer_meta, Layer)
```
afonsobspinto, line 85: suggestion (non-blocking): We can rename this method now. It no longer 'converts' it creates.

afonsobspinto (changes requested):
Overall looks good to me I just left some small change requests. 
As agreed internally I did not test this PR as I believe @ddelpiano will be doing that shortly.

afonsobspinto (dismissed):

backend/composer/services/cs_ingestion/helpers/anatomical_entities_helper.py
D-GopalKrishna: thanks! @afonsobspinto 👍

afonsobspinto (approved):

ddelpiano (dismissed):

ddelpiano (approved):

PR #289, opened 2024-07-10
CSCKAN-293 - adding missing migrations

D-GopalKrishna:
Missing migrations from - https://github.com/MetaCell/sckan-composer/pull/287


D-GopalKrishna:
closing since, the other PR from https://github.com/MetaCell/sckan-composer/pull/288 adds the migration missing here. 

cc @ddelpiano

PR #294, opened 2024-09-18
Feature/esckan-65 - get entities (origin, destination, vias) in the journey as part of serializer export

D-GopalKrishna:
- Added a new serializer field - `entities_journey`. This returns all the vias, destinations, and origins of that respective KS. 
- 
```
it is of the form:

entitiesJourney: EntitiesJourneyType[]

where EntitiesJourneyType  is;

export type EntitiesJourneyType = {
  origins: string[];
  vias: string[];
  destinations: string[];
};

```

- Each string in origin, vias, destination - is the name of that origin/via/destination.
- We get this information from the - [consolidated_paths](https://github.com/MetaCell/sckan-composer/blob/82c22ad198728d48113d554bfed688dc0b350f0f/backend/composer/services/graph_service.py#L91).

### DIFF
--- a/backend/composer/api/serializers.py
+++ b/backend/composer/api/serializers.py
@@ -516,6 +516,7 @@ class ConnectivityStatementSerializer(BaseConnectivityStatementSerializer):
     )
     available_transitions = serializers.SerializerMethodField()
     journey = serializers.SerializerMethodField()
+    entities_journey = serializers.SerializerMethodField()    
     statement_preview = serializers.SerializerMethodField()
     errors = serializers.SerializerMethodField()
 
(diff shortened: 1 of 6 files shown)


afonsobspinto (changes requested):
issue: Would it be possible to update [the journey tests](https://github.com/MetaCell/sckan-composer/blob/develop/backend/tests/test_journey.py) to match with the new format?

afonsobspinto (changes requested):
question: Would it make sense to run the [genapi.sh](https://github.com/MetaCell/sckan-composer/blob/develop/frontend/genapi.sh) to update the open api specification?

afonsobspinto (changes requested):
suggestion: It would be good if we also had the ids of the entities in the new entities journey property

ddelpiano (approved):
lgtm

PR #315, opened 2024-10-08
Feature/sckan 300

afonsobspinto:
Closes https://metacell.atlassian.net/browse/SCKAN-300

- Adds GraphRenderingState model (one to one with ConnectivityStatement) to store the serialized_graph json.
- Adds save icon to the graph diagram.
- Updates custom node to serialize and deserialize custom data.
- On click of save icon makes a patch request with the serialized_graph json data.
- Refactors out of graph component the algorithm to position the nodes and updates it to use the positions of the serialized_graph when possible:

https://github.com/user-attachments/assets/6de1d876-4d3b-43ca-a29d-9ad45fe76d38


backend/composer/models.py
```diff
@@ -713,6 +712,20 @@ class Meta:
...
+class GraphRenderingState(models.Model):
+    connectivity_statement = models.OneToOneField(
+        ConnectivityStatement,
+        on_delete=models.CASCADE,
+        related_name='graph_rendering_state',
+    )
+    serialized_graph = models.JSONField()  # Stores the serialized diagram model
+    created_at = models.DateTimeField(auto_now_add=True)
+    updated_at = models.DateTimeField(auto_now=True)
+    saved_by = models.ForeignKey(
```
ddelpiano, line 724: @afonsobspinto question, are the update_at and saved_by used to update the change in the notes? Thanks

afonsobspinto: No. At the moment they are not used but I added them anyway because I can see it being useful for something like that.

afonsobspinto: We can also have a note in the graph saying. Last saved by: {saved_by} at: {updated_at}

PR #317, opened 2024-10-11
Feature/sckan-312 - add flag to neurondm ingestion script to disable the overwrite of KSs

D-GopalKrishna:
Jira Link - [SCKAN-312](https://metacell.atlassian.net/browse/SCKAN-312)

Task description:
1. add logic to disable overwrite of statements when the flag --disable_overwrite is used. 
2. Renamed overwritable_statements to overwritable_and_new_statements (I think it better represents the statements types)

More on the previous vs updated logic:

After the changes the definition of `overwritable_and_new_statements` changes. Whether the statements are overwritable or not depends on two utils - `has_invalid_sentence` and `has_invalid_sentence`. 

(Previously and now) if the statement/sentence didn't exist we say it's not invalid. And therefore overwritable (new statement/sentence). This logic stays the same. 

Previously - if the statement/sentence exists, and was overwritable (depending on if it is invalid/exported/compose_now) - then it is not invalid, and hence overwritable. 
Now (after the change) - WE ONLY CHANGE THIS LOGIC above. This check only happens --disable_overwrite flag is not passed. If the flag is passed - we do not check this, and say the statement/sentence is invalid and hence - not overwritable. 

Short demo video to showcase the same. 

https://github.com/user-attachments/assets/59185134-f812-47a9-962a-f6ef4e0ee510

demo after the logic is enabled. 

https://github.com/user-attachments/assets/5bb56a06-35f8-4fd8-b838-9762412991dd

[SCKAN-312]: https://metacell.atlassian.net/browse/SCKAN-312?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


backend/composer/services/cs_ingestion/helpers/overwritable_helper.py
```diff
@@ -10,25 +10,35 @@
 logger_service = LoggerService()
 
 
-def get_overwritable_statements(statements_list: List[Dict[str, Any]]) -> List[Dict[str, Any]]:
-    overwritable_statements = [
+def get_overwritable_and_new_statements(statements_list: List[Dict[str, Any]], disable_overwrite: bool=False) -> List[Dict[str, Any]]:
+    overwritable_and_new_statements = [
         statement for statement in statements_list
-        if not has_invalid_sentence(statement) and not has_invalid_statement(statement)
+        if not has_invalid_sentence(statement, disable_overwrite) and not has_invalid_statement(statement, disable_overwrite)
```
D-GopalKrishna, line 16: Because it depends on both the utils here. As you can see after running the script (from the second video). It updates the Connectivity Statement to len 426 (from 423), and Sentences to be 466 (from 463). Essentially adding 3. 

here Sentences are back (3 deleted, 3 ingested). However the Statements are not - I deleted 43, ingested Connectivity statements are only 3, (40 more are not ingestable). 

@afonsobspinto what do you think? should we update the logic here? (Or is it the case that the test case I used was a peculiar/non-realistic one).

ddelpiano (approved):
lgtm

backend/composer/services/cs_ingestion/helpers/overwritable_helper.py
afonsobspinto, line 16: suggestion: I think we should rename the has_invalid_x to is_new_or_overwritable_x (this will reverse the logic and remove the nots, making it more readable). 

Whoever came up with the has_invalid name should rethink their life decisions (it was me 😛)

backend/composer/services/cs_ingestion/cs_ingestion_services.py
```diff
@@ -14,10 +14,10 @@
 logger_service = LoggerService()
 
 
-def ingest_statements(update_upstream=False, update_anatomical_entities=False, should_overwrite=True, full_imports=[], label_imports=[]):
+def ingest_statements(update_upstream=False, update_anatomical_entities=False, disable_overwrite=False, full_imports=[], label_imports=[]):
     statements_list = get_statements_from_neurondm(full_imports=full_imports, label_imports=label_imports, logger_service_param=logger_service)
-    overridable_statements = get_overwritable_statements(statements_list)
-    statements = validate_statements(overridable_statements, update_anatomical_entities)
+    overridable_and_new_statements = get_overwritable_and_new_statements(statements_list, disable_overwrite)
```
afonsobspinto, line 19: This is a better name, thanks 🙏

backend/composer/services/cs_ingestion/helpers/overwritable_helper.py
afonsobspinto: The 40 statements were not ingested because the sentences were not in COMPOSE_NOW? If that's the case it sounds correct to me  (@ddelpiano can you double check here, what are the conditions to a statement be ingested when the disable override is set to true)

afonsobspinto (changes requested):

backend/composer/services/cs_ingestion/cs_ingestion_services.py
```diff
@@ -14,10 +14,10 @@
 logger_service = LoggerService()
 
 
-def ingest_statements(update_upstream=False, update_anatomical_entities=False, should_overwrite=True, full_imports=[], label_imports=[]):
+def ingest_statements(update_upstream=False, update_anatomical_entities=False, disable_overwrite=False, full_imports=[], label_imports=[]):
```
afonsobspinto, line 17: issue: Same as in the other PR, maybe it makes more sense to change here given the target of the PR

backend/composer/services/cs_ingestion/helpers/overwritable_helper.py
D-GopalKrishna: Great point Afonso. It helped a lot.

afonsobspinto (dismissed):

backend/composer/services/cs_ingestion/helpers/overwritable_helper.py
afonsobspinto: Thanks. Reads much better now 💯

ddelpiano:
@D-GopalKrishna both 311 and 312 are good for me, the only thing I would like to see are tests for both the features. Thanks a lot and great work :)

D-GopalKrishna:
NOTE: added tests. This increases the time to run the tests significantly - from literally under 1s to more than 400s.

backend/tests/test_ingest_statements.py
```diff
@@ -0,0 +1,90 @@
+from composer.services.cs_ingestion.cs_ingestion_services import ingest_statements
+from composer.models import ConnectivityStatement
+import unittest
+from composer.services.cs_ingestion.neurondm_script import log_error
+from composer.enums import CSState
+from django.db.models import Q
+
+
+class TestIngestStatements(unittest.TestCase):
```
afonsobspinto, line 9: suggestion (non-blocking): Use django.TestCase instead of unittest.TestCase. As mentioned privately, as a rule of thumb, when the tests require database access I prefer to use django's TestCase instead of unittest.

backend/tests/test_ingest_statements.py
```diff
@@ -0,0 +1,90 @@
...
+    """
+    NOTE:
+    This test depends on the directory system of the Scicrunch neurondm - here - https://raw.githubusercontent.com/SciCrunch/NIF-Ontology/neurons/**/*.ttl
+    Check more details in neurondm_script.py
+    """
+
+    def flush_connectivity_statements(self):
+        ConnectivityStatement.objects.all().delete()
+
+    def test_full_and_label_import(self):
```
afonsobspinto, line 19: suggestion (non-blocking): This test seems to be doing a lot of ingestions. Would it make sense to split in different tests? One for each?

backend/tests/test_ingest_statements.py
```diff
@@ -0,0 +1,90 @@
...
+        self.assertNotEqual(ConnectivityStatement.objects.count(), 0)
+
+        self.flush_connectivity_statements()
+
+        # check for both valid full and label imports
+        ingest_statements(
+            full_imports=['sparc-nlp'],
+            label_imports=['apinatomy-neuron-populations']
+        )
+        self.assertNotEqual(ConnectivityStatement.objects.count(), 0)
```
afonsobspinto, line 45: suggestion (non-blocking): I don't think "at least 1 statement was ingested" is equivalent to "the ingestion is working as expected". It's a good indication but not a guarantee. My proposal is to create a new card to improve the tests by, for example, mocking the SciCrunch data files and set the exact number of statements we expect to be ingested on each test

afonsobspinto (dismissed):

backend/tests/test_ingest_statements.py
D-GopalKrishna: Updated! thanks :)

backend/tests/test_ingest_statements.py
D-GopalKrishna: Done, makes it better yes thanks!

backend/tests/test_ingest_statements.py
D-GopalKrishna: I agree that it doesn't mean - "the ingestion is working as expected", but this test is to test - given the right set of full and label imports, the `ingest_statements()` ingests the statements. If the labels are wrong inputs, then it doesn't import. 

Question: @afonsobspinto - do you think - the statement count would be a constant? or how often would it be changed from the Scicrunch side - in the official source?

D-GopalKrishna: Had a chat with @ddelpiano! Will add a card to tech debt for further tests on checking counts.

afonsobspinto (approved):

ddelpiano (approved):

PR #342, opened 2024-10-23
check ownership before performing the update

Salam-Dalloul:


frontend/src/helpers/settings.ts
```diff
@@ -28,3 +28,8 @@ export const sentenceStateToColor: SentenceStateToColor = {
 export const OriginsGroupLabel = "Origins";
 export const ViasGroupLabel = "Vias";
 export const DestinationsGroupLabel = "Destinations";
+
+export const ChangeRequestStatus = {
```
afonsobspinto, line 32: Note: This was a good idea, you should use this const instead of the hardcoded values

ddelpiano (changes requested):
@Salam-Dalloul for the console messages and ts-ignore I will open a separate card to look at them, could you please address the issue noticed in the StatementService.ts? after that we can merge this and we'll address the other problems separately. Thanks!

afonsobspinto (commented):
The save after I request to be the owner doesn't seem to be working:
https://github.com/user-attachments/assets/cde251fb-9ed1-405c-9e9d-50a946c54abf


I'm getting:
{"owner_id":["A valid integer is required."]}

The owner id seems to be filled incorrectly with the complete statement
image

frontend/src/Pages/SentenceDetails.tsx
```diff
@@ -176,22 +186,6 @@ const SentencesDetails = () => {
           setConnectivityStatements(
             sentence.connectivity_statements.sort((a, b) => a.id - b.id),
           );
-          if (
```
afonsobspinto, line 179: I took the decision of removing the initial alert from the statement because in that page the user can still try to edit the statement (even if they are not the owner) which would trigger the appearance of the ownership alert. That doesn't seem to be the case for the sentences. In other words, with this change, I don't believe it's possible to re-take the ownership of a sentence. I propose that we don't remove this initial prompt from the sentences.

ddelpiano: discussed with @Salam-Dalloul

PR #354, opened 2024-11-07
Feature/sckan-273 

D-GopalKrishna:
Jira ticket - [SCKAN-273](https://metacell.atlassian.net/browse/SCKAN-273)

About the issue - of statements getting converted from exported to invalid and vice versa. 

Previously:
`ConnectivityStatement.objects.filter(reference_uri=reference_uri).update(**defaults)`

IMO - we should not update the [default state](https://github.com/MetaCell/sckan-composer/blob/79c5ff869fcf3e39f53e8345e50a1acb739a933f/backend/composer/services/cs_ingestion/helpers/statement_helper.py#L22) here (since the default state is [EXPORTED](https://github.com/MetaCell/sckan-composer/blob/79c5ff869fcf3e39f53e8345e50a1acb739a933f/backend/composer/services/cs_ingestion/helpers/statement_helper.py#L22)). So the problem we were facing - happened because if the statement had any changes, then - it was getting [converted to EXPORTED](https://github.com/MetaCell/sckan-composer/blob/79c5ff869fcf3e39f53e8345e50a1acb739a933f/backend/composer/services/cs_ingestion/helpers/statement_helper.py#L40).

<hr />

This alone should fix the issue. Since now - it is not getting converted to EXPORTED, and hence would not be needing the conversion back to INVALID. However following is a `good to have`.

i.e. 
`connectivity_statement = ConnectivityStatement.objects.filter(reference_uri=reference_uri).first()`

instead of 

```
fields_to_refresh = [field for field in defaults.keys() if field != 'state']
connectivity_statement.refresh_from_db(fields=fields_to_refresh)
```

since it doesn't get the latest for the state field. Which later caused errors (before the above fix) [here](https://github.com/MetaCell/sckan-composer/blob/79c5ff869fcf3e39f53e8345e50a1acb739a933f/backend/composer/services/cs_ingestion/helpers/statement_helper.py#L50). 

Example of bad update before:
Once the statement was converted to EXPORTED state due to defaults, [line](https://github.com/MetaCell/sckan-composer/blob/79c5ff869fcf3e39f53e8345e50a1acb739a933f/backend/composer/services/cs_ingestion/helpers/statement_helper.py#L49) still thinks it is in invalid state (and hence doesn't move it to INVALID - but will convert this very statement to INVALID the next time we run the script). However, we did [update](https://github.com/MetaCell/sckan-composer/blob/79c5ff869fcf3e39f53e8345e50a1acb739a933f/backend/composer/services/cs_ingestion/helpers/statement_helper.py#L40) and [fetched the latest](https://github.com/MetaCell/sckan-composer/blob/79c5ff869fcf3e39f53e8345e50a1acb739a933f/backend/composer/services/cs_ingestion/helpers/statement_helper.py#L42) the before. 

[SCKAN-273]: https://metacell.atlassian.net/browse/SCKAN-273?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/composer/models.py
+++ b/backend/composer/models.py
@@ -634,7 +634,10 @@ class ConnectivityStatement(models.Model):
 
     @transition(
         field=state,
-        source=CSState.NPO_APPROVED,
+        source=[
+            CSState.NPO_APPROVED,
+            CSState.INVALID
+        ],
         target=CSState.EXPORTED,
         conditions=[ConnectivityStatementStateService.is_valid],
         permission=ConnectivityStatementStateService.has_permission_to_transition_to_exported,
--- a/backend/composer/services/cs_ingestion/helpers/statement_helper.py
+++ b/backend/composer/services/cs_ingestion/helpers/statement_helper.py
@@ -12,7 +12,7 @@ from composer.services.cs_ingestion.helpers.common_helpers import LABEL, VALIDAT
 from composer.services.cs_ingestion.helpers.getters import get_sex, get_circuit_type, get_functional_circuit_role, \
     get_phenotype, get_projection_phenotype
 from composer.services.cs_ingestion.helpers.notes_helper import do_transition_to_invalid_with_note, create_invalid_note, \
-    add_ingestion_system_note
+    add_ingestion_system_note, do_transition_to_exported
 from composer.services.cs_ingestion.models import ValidationErrors
 
 
@@ -50,6 +50,9 @@ def create_or_update_connectivity_statement(statement: Dict, sentence: Sentence,
             do_transition_to_invalid_with_note(connectivity_statement, error_message)
         else:
             create_invalid_note(connectivity_statement, error_message)
+    else:
+        if connectivity_statement.state != CSState.EXPORTED:
+            do_transition_to_exported(connectivity_statement)
 
     update_many_to_many_fields(connectivity_statement, statement, update_anatomical_entities)
     statement[STATE] = connectivity_statement.state
--- a/backend/composer/services/cs_ingestion/helpers/statement_helper.py
+++ b/backend/composer/services/cs_ingestion/helpers/statement_helper.py
@@ -35,11 +35,11 @@ def create_or_update_connectivity_statement(statement: Dict, sentence: Sentence,
         reference_uri=reference_uri,
         defaults=defaults
     )
-    if not created:
+    if not created:        
         if has_changes(connectivity_statement, statement, defaults):
-            ConnectivityStatement.objects.filter(reference_uri=reference_uri).update(**defaults)
-            fields_to_refresh = [field for field in defaults.keys() if field != 'state']
-            connectivity_statement.refresh_from_db(fields=fields_to_refresh)
+            defaults_without_state = {field: value for field, value in defaults.items() if field != 'state'}
+            ConnectivityStatement.objects.filter(reference_uri=reference_uri).update(**defaults_without_state)
+            connectivity_statement = ConnectivityStatement.objects.filter(reference_uri=reference_uri).first()
             add_ingestion_system_note(connectivity_statement)
 
     validation_errors = statement.get(VALIDATION_ERRORS, ValidationErrors())
--- a/backend/composer/services/cs_ingestion/helpers/statement_helper.py
+++ b/backend/composer/services/cs_ingestion/helpers/statement_helper.py
@@ -39,7 +39,7 @@ def create_or_update_connectivity_statement(statement: Dict, sentence: Sentence,
         if has_changes(connectivity_statement, statement, defaults):
             defaults_without_state = {field: value for field, value in defaults.items() if field != 'state'}
             ConnectivityStatement.objects.filter(reference_uri=reference_uri).update(**defaults_without_state)
-            connectivity_statement = ConnectivityStatement.objects.filter(reference_uri=reference_uri).first()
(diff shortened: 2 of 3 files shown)


backend/composer/services/cs_ingestion/helpers/statement_helper.py
```diff
@@ -35,11 +35,11 @@ def create_or_update_connectivity_statement(statement: Dict, sentence: Sentence,
...
         defaults=defaults
     )
-    if not created:
+    if not created:        
         if has_changes(connectivity_statement, statement, defaults):
-            ConnectivityStatement.objects.filter(reference_uri=reference_uri).update(**defaults)
-            fields_to_refresh = [field for field in defaults.keys() if field != 'state']
-            connectivity_statement.refresh_from_db(fields=fields_to_refresh)
+            defaults_without_state = {field: value for field, value in defaults.items() if field != 'state'}
+            ConnectivityStatement.objects.filter(reference_uri=reference_uri).update(**defaults_without_state)
```
D-GopalKrishna, line 41: @afonsobspinto - keeping the filter instead of the get here - since we cannot update a get
`AttributeError: 'ConnectivityStatement' object has no attribute 'update'`

also doing `connectivity_statement = ConnectivityStatement.objects.filter(reference_uri=reference_uri).first()` since, we get 

```
ConnectivityStatement.objects.filter(reference_uri=reference_uri).update(**defaults_without_state)
>>> 1
```
as return value after the update.

afonsobspinto:
Just to clarify a little bit the original problem.

As you mentioned, if an existent statement had changes it would transition to exported. This is in theory non problematic because later there's a set of instructions responsible to move it to (invalid)[https://github.com/MetaCell/sckan-composer/blob/79c5ff869fcf3e39f53e8345e50a1acb739a933f/backend/composer/services/cs_ingestion/helpers/statement_helper.py#L49]. However the state of the statement was [**not** refreshed](https://github.com/MetaCell/sckan-composer/blob/79c5ff869fcf3e39f53e8345e50a1acb739a933f/backend/composer/services/cs_ingestion/helpers/statement_helper.py#L41) and that's where the real problem lies. 

Here's an example of the problem:

Statement A exists and is Invalid
Statement A moves to Exported by default due to being part of the ingestion
We check if statements with errors are in Invalid state, statement A is still with state set as Invalid so no transition is applied.
We end up with Statement A in Exported even tho it should be in Invalid

backend/composer/services/cs_ingestion/helpers/statement_helper.py
afonsobspinto: I believe we could do:

            ```
            ConnectivityStatement.objects.filter(reference_uri=reference_uri).update(**defaults_without_state)
            connectivity_statement.refresh_from_db()
            ```

afonsobspinto (approved):

afonsobspinto (changes requested):
I understand that the solution proposed is to not do the initial transition to Exported.
Let's see the possible scenarios:
1. Statement A is in Exported and has no problems ✅ 
2. Statement A is in Exported and has problems, then it's moved to Invalid by the transition check ✅ 
3. Statement A is in Invalid and has problems, a new note is added, no transition is necessary ✅ 
4. Statement A is in Invalid and has no problems. It should be moved to exported but currently I don't think it's. ❌ 

I think we need to add something, after the update without the state, on the lines of:
```
        if connectivity_statement.state != CSState.EXPORTED:
            # Use the FSM transition method
            connectivity_statement.exported()
            connectivity_statement.save()
            
            ```

File: backend/composer/services/cs_ingestion/helpers/statement_helper.py (hunk 1 above)
Comment (D-GopalKrishna) [reply backend/composer/services/cs_ingestion/helpers/statement_helper.py:41]: Do you mean the following?

```
fields_to_refresh = [field for field in defaults.keys() if field != 'state']
connectivity_statement.refresh_from_db(fields=fields_to_refresh)
```


since the `connectivity_statement.refresh_from_db()` would lead to `AttributeError: Direct state modification is not allowed`.

File: backend/composer/services/cs_ingestion/helpers/statement_helper.py [hunk 2]
```diff
@@ -50,6 +50,9 @@ def create_or_update_connectivity_statement(statement: Dict, sentence: Sentence,
             do_transition_to_invalid_with_note(connectivity_statement, error_message)
         else:
             create_invalid_note(connectivity_statement, error_message)
+    else:
+        if connectivity_statement.state != CSState.EXPORTED:
+            do_transition_to_exported(connectivity_statement)
```
D-GopalKrishna, line 55: move to EXPORTED state (default) if not in EXPORTED state - if there's no validation errors. 

Reason to include this: (credit thanks - @afonsobspinto : ) )

If a statement was in INVALID state in the past; but if it has no validation errors when running it again in the future, then it should to be moved to EXPORTED state. 

To cover the fourth edge case here - https://github.com/MetaCell/sckan-composer/pull/354#pullrequestreview-2427647132

afonsobspinto (dismissed):

backend/composer/models.py
```diff
@@ -634,7 +634,10 @@ def npo_approved(self, *args, **kwargs):
 
     @transition(
         field=state,
-        source=CSState.NPO_APPROVED,
+        source=[
```
D-GopalKrishna, line 637: NOTE: allow to export from invalid.

PR #355, opened 2024-11-11
Feature/sckan 300 b

afonsobspinto:
closes https://metacell.atlassian.net/browse/SCKAN-300

- Adds signals on origin, vias and destination changes that remove the graph state if it exists
- Updates statement viewset to return updated entity 
- Updates frontend to setStatement on origin changes and refresh statement on vias/destination changes:

https://github.com/user-attachments/assets/af235969-2ae9-45c2-95d3-48a4d3b1f6b1

### DIFF
--- a/frontend/src/components/Forms/StatementForm.tsx
+++ b/frontend/src/components/Forms/StatementForm.tsx
@@ -252,12 +252,13 @@ const StatementForm = (props: any) => {
           onUpdate: async (selectedOption: string, formId: string) => {
             const viaIndex = getConnectionId(formId, statement.vias);
             const typeOption = selectedOption as TypeB60Enum;
-            
+
             if (viaIndex) {
               try {
                 await api.composerViaPartialUpdate(viaIndex, {
                   type: typeOption,
                 });
+                refreshStatement()
                 return ChangeRequestStatus.SAVED;
               } catch (error) {
                 return checkOwnership(
@@ -266,6 +267,7 @@ const StatementForm = (props: any) => {
                     await api.composerViaPartialUpdate(viaIndex, {
                       type: typeOption,
                     });
+                    refreshStatement()
                     return ChangeRequestStatus.SAVED;
                   },
                   () => {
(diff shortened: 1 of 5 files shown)


frontend/src/components/Forms/StatementForm.tsx
```diff
@@ -446,13 +450,13 @@ const StatementForm = (props: any) => {
           isPathBuilderComponent: true,
           InputIcon: DestinationIcon,
           onUpdate: async (selectedOption: string, formId: string) => {
-            const viaIndex = getConnectionId(formId, statement?.destinations);
+            const destinationIndex = getConnectionId(formId, statement?.destinations);
             const typeOption = selectedOption as TypeC11Enum;
-            if (viaIndex) {
```
afonsobspinto, line 451: I find it strange that this is not wrapped in a checkOwnership function. I suspect that the UI won't behave correctly when trying to change the type of a destination from a connectivity statement that I don't own. I didn't test this tho. cc @ddelpiano @Salam-Dalloul

frontend/src/components/Forms/StatementForm.tsx
```diff
@@ -446,13 +450,13 @@ const StatementForm = (props: any) => {
...
-              api
-                .composerDestinationPartialUpdate(viaIndex, {
-                  type: typeOption,
-                })
+            if (destinationIndex) {
+              await api.composerDestinationPartialUpdate(destinationIndex, {
```
afonsobspinto, line 456: I find it strange that this is not wrapped in a checkOwnership function. I suspect that the UI won't behave correctly when trying to change the type of a destination from a connectivity statement that I don't own. I didn't test this tho. cc @ddelpiano @Salam-Dalloul

ddelpiano: Good catch, I will open a card for @Salam-Dalloul for this. Thanks

ddelpiano: this has been fixed as well

ddelpiano (approved):

PR #361, opened 2024-11-18
Frontend implementation for Alert system

Salam-Dalloul:


frontend/src/components/DistillationTab/StatementAlertsAccordion.tsx
```diff
@@ -0,0 +1,352 @@
...
+          ...(statement.statement_alerts || []),
+          res,
+        ];
+        const updatedStatement = { ...statement, statement_alerts: updatedAlerts };
+        setActiveTypes([...activeTypes, typeId]);
+        setStatement(updatedStatement);
+        const newIndex = updatedAlerts.length - 1;
+        setOpenFormIndex(newIndex);
+        setTimeout(() => {
+          const textArea = document.querySelectorAll(`#root_statement_alerts_0_text`);
```
afonsobspinto, line 77: suggestion: I would prefer that we don't rely on a hardcoded id to focus, instead maybe we can have a reference to the statement alerts element and use it here

afonsobspinto (approved):

afonsobspinto (changes requested):
As discussed privately, if it's easy I would prefer the useEffect implementation instead of the setTimeout

PR #370, opened 2024-12-05
Feature/sckan 326

afonsobspinto:
Closes https://metacell.atlassian.net/browse/SCKAN-326

- Updates graph rendering strategy:
1- We collect the information about the entities, this tell us how they connect with each other and what type (Via, Origin, Destination) they are
2 - We go bottom -> up from afferent terminals (creating and storing nodes and links along the way)
3 - We go bottom -> up from non afferent terminals (reusing existent nodes and links or creating new ones along the way, the list of created nodes and links is reversed so they are added to the engine in the 'natural' order

- Updates widgets styles and ports

https://github.com/user-attachments/assets/1feaf304-ad4e-4803-87f2-9c561446e66b


ddelpiano (approved):
lgtm

PR #379, opened 2024-12-10
SCKAN-326 feat: Add disconnected nodes to graph

afonsobspinto:
Closes https://metacell.atlassian.net/browse/SCKAN-326

- Adds nodes without a  (direct or indirect)  path to a destination as disconnected nodes

image


ddelpiano (approved):
lgtm

PR #381, opened 2024-12-10
Feature/sckan-324 -  journey computation

D-GopalKrishna:
Jira Link - [SCKAN-324](https://metacell.atlassian.net/browse/SCKAN-324)

[ONLY BONUS DEVELOPMENT] - Others are not part of this PR and are explained below:
a. I didn't not see any rows that went empty. 
b. export of statement preview is not implemented in this PR. Perhaps we can change the name of the card @ddelpiano ?

Task Implementation description:
Refactor the journey computation
1. persist the journey in the db
2. re-compute this everytime we change the entities in path builder are changed - adapting PATCH request to save connectivity statements too. 

Demo

https://github.com/user-attachments/assets/55fd6785-6369-409b-a4a4-3329e55a71a4

[SCKAN-324]: https://metacell.atlassian.net/browse/SCKAN-324?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


backend/composer/models.py
```diff
@@ -566,6 +566,8 @@ class ConnectivityStatement(models.Model):
     )
     created_date = models.DateTimeField(auto_now_add=True, db_index=True)
     modified_date = models.DateTimeField(auto_now=True, db_index=True)
+    journey_description = models.JSONField(null=True, blank=True)
```
afonsobspinto, line 569: Suggestion: My preference would be to have just 1 attribute since they need to always be in sync.

backend/composer/api/views.py
```diff
@@ -531,6 +538,14 @@ class DestinationViewSet(viewsets.ModelViewSet):
     permission_classes = [IsOwnerOfConnectivityStatementOrReadOnly]
     filterset_class = DestinationFilter
 
+    def partial_update(self, request, *args, **kwargs):
```
afonsobspinto, line 541: issue: I understand the idea but I don't believe partial_update is triggered for all the cases where we would need to recompute the journey (for example, in the case we delete a complete Via).

We already have django signals in place that cover the exact places we need: destination_changed, via_changed, destination_from_entities_changed, destination_anatomical_entities_changed, via_from_entities_changed, via_anatomical_entities_changed, connectivity_statement_origins_changed.

 My recommendation would be to run the 
        `cs.save(update_fields=["journey_description", "journey_entities"])`
or equivalent

backend/composer/migrations/0064_auto_20241208_2313.py
```diff
@@ -0,0 +1,24 @@
+# Generated by Django 4.1.4 on 2024-12-08 22:13
+
+from django.db import migrations
+from composer.services.graph_service import compile_journey
+
+
+def update_journey_fields(apps, schema_editor):
```
afonsobspinto, line 7: 👍

backend/composer/services/graph_service.py
```diff
@@ -157,12 +157,12 @@ def compile_journey(connectivity_statement) -> dict:
...
+    Origin = apps.get_model('composer', 'AnatomicalEntity')
 
-    origins = list(connectivity_statement.origins.all())
-
-    vias = list(Via.objects.filter(connectivity_statement=connectivity_statement))
-    destinations = list(Destination.objects.filter(connectivity_statement=connectivity_statement))
-
+    vias = list(Via.objects.filter(connectivity_statement__id=connectivity_statement.id))
+    destinations = list(Destination.objects.filter(connectivity_statement__id=connectivity_statement.id))
+    origins = list(Origin.objects.filter(origins_relations__id=connectivity_statement.id).distinct())
```
afonsobspinto, line 164: question: Why did we change from:
    origins = list(connectivity_statement.origins.all())
    to
     origins = list(Origin.objects.filter(origins_relations__id=connectivity_statement.id).distinct())

afonsobspinto (changes requested):

backend/composer/api/views.py
D-GopalKrishna: yes thank you @afonsobspinto ! I made the change.

backend/composer/services/graph_service.py
D-GopalKrishna: Reason is here in the code - https://github.com/MetaCell/sckan-composer/blob/feature/SCKAN-324/backend/composer/services/graph_service.py#L24

It says:
```
Exception has occurred: AttributeError       (note: full exception trace is shown but execution is paused at: _run_module_as_main)
'AnatomicalEntity' object has no attribute 'name'
  File "/home/dgk/metacell/sckan-composer/backend/composer/services/graph_service.py", line 24, in generate_paths
    paths.extend(create_paths_from_origin(origin, vias, destinations, [(str(origin.id), origin.name, 0)], destination_layer))
  File "/home/dgk/metacell/sckan-composer/backend/composer/services/graph_service.py", line 166, in compile_journey
    all_paths2 = generate_paths(origins, vias, destinations)
  File "/home/dgk/metacell/sckan-composer/backend/composer/migrations/0067_auto_20241218_0928.py", line 11, in update_journey_fields
    cs.journey_path = compile_journey(cs)
  File "/home/dgk/miniconda3/envs/composer/lib/python3.9/site-packages/django/db/migrations/operations/special.py", line 193, in database_forwards
    self.code(from_state.apps, schema_editor)
  File "/home/dgk/miniconda3/envs/composer/lib/python3.9/site-packages/django/db/migrations/migration.py", line 130, in apply
    operation.database_forwards(
  File "/home/dgk/miniconda3/envs/composer/lib/python3.9/site-packages/django/db/migrations/executor.py", line 252, in apply_migration
    state = migration.apply(state, schema_editor)
  File "/home/dgk/miniconda3/envs/composer/lib/python3.9/site-packages/django/db/migrations/executor.py", line 167, in _migrate_all_forwards
    state = self.apply_migration(
  File "/home/dgk/miniconda3/envs/composer/lib/python3.9/site-packages/django/db/migrations/executor.py", line 135, in migrate
    state = self._migrate_all_forwards(
  File "/home/dgk/miniconda3/envs/composer/lib/python3.9/site-packages/django/core/management/commands/migrate.py", line 349, in handle
    post_migrate_state = executor.migrate(
  File "/home/dgk/miniconda3/envs/composer/lib/python3.9/site-packages/django/core/management/base.py", line 96, in wrapped
    res = handle_func(*args, **kwargs)
  File "/home/dgk/miniconda3/envs/composer/lib/python3.9/site-packages/django/core/management/base.py", line 448, in execute
    output = self.handle(*args, **options)
  File "/home/dgk/miniconda3/envs/composer/lib/python3.9/site-packages/django/core/management/base.py", line 402, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/home/dgk/miniconda3/envs/composer/lib/python3.9/site-packages/django/core/management/__init__.py", line 440, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/home/dgk/miniconda3/envs/composer/lib/python3.9/site-packages/django/core/management/__init__.py", line 446, in execute_from_command_line
    utility.execute()
  File "/home/dgk/metacell/sckan-composer/backend/manage.py", line 18, in main
    execute_from_command_line(sys.argv)
  File "/home/dgk/metacell/sckan-composer/backend/manage.py", line 22, in <module>
    main()
  File "/home/dgk/miniconda3/envs/composer/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/dgk/miniconda3/envs/composer/lib/python3.9/runpy.py", line 197, in _run_module_as_main (Current frame)
    return _run_code(code, main_globals, None,
AttributeError: 'AnatomicalEntity' object has no attribute 'name'
```

backend/composer/models.py
```diff
@@ -716,6 +717,9 @@ def save(self, *args, **kwargs):
             self.reference_uri = create_reference_uri(self.pk)
             self.save(update_fields=["reference_uri"])
 
+        # recompute - journey
+        self.journey_path = compile_journey(self)
+        super().save(update_fields=["journey_path"])
```
afonsobspinto, line 722: Question: Do we need to update the journey on connectivity statement changes? 
In other words, what change affects the journey that is not capture in the signals?

Computing the journey is somewhat expensive so I would prefer that we make sure we only do it when we really need to

backend/composer/services/graph_service.py
afonsobspinto: I tried to access the name of an origin with the original code and I was able to:
image

Can you explain with a bit more detail in which conditions the error you shown occurs?

D-GopalKrishna: @afonsobspinto did you do the check with the API call?

This happens only when we do the migration - possibly something with - `apps.get_model('composer', 'ConnectivityStatement')` in the migration file (ref - [0067 auto migration file](https://github.com/MetaCell/sckan-composer/blob/feature/SCKAN-324/backend/composer/migrations/0067_auto_20241218_0928.py#L8)). Also if you notice this is only for the `@properties` and not for the fields. 
So if you put a breakpoint to the above mentioned code, and then run the migration to db  - this should cause the error.

backend/composer/models.py
D-GopalKrishna: Interesting point @afonsobspinto, I agree - so we just do the computation of the journey path only in the signals then. When the entities are changed, and therefore the journey should also be changed.  

Signals should cover the requirements for us. thanks for pointing this!
Made the change.

backend/composer/services/graph_service.py
afonsobspinto: Understood, during migrations django uses historical models that don't include python level constructs like @property.
I think the new instruction works, but I would remove the .distinct(), I don't think it's needed in our context

afonsobspinto (changes requested):
Would it be possible to also change the admin page to not show or keep the journey path as read only?

image

backend/composer/services/graph_service.py
```diff
@@ -157,17 +155,55 @@ def compile_journey(connectivity_statement) -> dict:
...
+        node[2] == 0 or path.index(node) == len(path) - 1) else (
+        node[1].replace(JOURNEY_DELIMITER, ', '), node[2])) for node in path] for path in consolidated_paths]
+    return journey_paths
+
+
+def build_journey_description(consolidated_paths, connectivity_statement):
+    journey_paths = get_journey_path_from_consolidated_paths(
+        consolidated_paths)
+    Via = apps.get_model('composer', 'Via')
+    vias = list(Via.objects.filter(
```
afonsobspinto, line 182: question: Why do we need to fetch the vias here? can't we rely entirely on the new journey paths attribute created?

afonsobspinto (changes requested):

backend/composer/services/graph_service.py
```diff
@@ -157,17 +155,55 @@ def compile_journey(connectivity_statement) -> dict:
...
+            sentence = f"from {origin_names} to {destination_names}"
 
+        journey_descriptions.append(sentence)
+    return journey_descriptions
+
+
+def build_journey_entities(consolidated_paths, connectivity_statement):
     entities = []
+    Via = apps.get_model('composer', 'Via')
+    vias = list(Via.objects.filter(
```
afonsobspinto, line 204: Same question as https://github.com/MetaCell/sckan-composer/pull/381/files#r1904313466

backend/composer/services/graph_service.py
D-GopalKrishna: Before removing it, I want to confirm again with you if you are okay to remove the `distinct()` here, the reason I had that is because unlike vias and destinations - which are FK, origins is a M2M.

afonsobspinto: The ManyToManyField in Django automatically ensures that the relationship between two models is unique. In other words, we know that we have a unique pair of ConnectivityStatement and AnatomicalEntity. Here's a simple test:

image

afonsobspinto: With that said, if we have the time I would apply the same suggestion I mentioned here:
https://github.com/MetaCell/sckan-composer/pull/404#discussion_r1915479028

backend/composer/services/graph_service.py
afonsobspinto: I created this card to handle this at a later stage:
https://metacell.atlassian.net/browse/SCKAN-390

backend/composer/services/graph_service.py
```diff
@@ -157,46 +155,67 @@ def compile_journey(connectivity_statement) -> dict:
...
+        if via_names:
+            sentence = f"from {origin_names} to {destination_names} via {via_names}"
+        else:
+            sentence = f"from {origin_names} to {destination_names}"
 
+        journey_descriptions.append(sentence)
+    return journey_descriptions
+
+
+def build_journey_entities(consolidated_paths, connectivity_statement):
```
afonsobspinto, line 200: With the latest changes connectivity_statement is no longer needed as a parameter here

backend/composer/admin.py
```diff
@@ -231,7 +231,7 @@ class ConnectivityStatementAdmin(
     list_per_page = 10
     # The name of one or more FSMFields on the model to transition
     fsm_field = ("state",)
-    readonly_fields = ("state",)
+    readonly_fields = ("state", "journey_path")
```
afonsobspinto, line 234: Since journey_path is a complex object and not the string I think it might make more sense to actually exclude it instead of having it as read-only

afonsobspinto (dismissed):

backend/composer/migrations/0067_auto_20241218_0928.py
```diff
@@ -0,0 +1,21 @@
...
+    ConnectivityStatement = apps.get_model('composer', 'ConnectivityStatement')
+    for cs in ConnectivityStatement.objects.all():
+        cs.journey_path = get_compile_journey_for_migration(cs)
+        cs.save(update_fields=["journey_path"])
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("composer", "0066_connectivitystatement_journey_path"),
```
afonsobspinto, line 16: issue: This migration does not exist

afonsobspinto (changes requested):

afonsobspinto (approved):

PR #404, opened 2025-01-05
Feature/SCKAN-364 - Add statement preview row to the export + persist statement preview suffix and prefix

D-GopalKrishna:
Jira link - https://metacell.atlassian.net/browse/SCKAN-364
NOTE THE BASE branch is `feature/SCKAN-324`.

Task description
- add statement_prefix and statement_suffix
- add migration to add values for already existing Connectivity Statements
- add get_suffix_for_statement_preview and get_prefix_for_statement_preview to statement service for separating logic for suffix and prefix creation
- handle the saving/updating of the statement preview using signals 

Why does the signal have `updated_fields`?
-> Because signal can be called for any changes that are not directly related to the fields updated - like the image below. And this logic makes sure - that we update only either when the m2m changes or direct field changes occur. 
image

In Export's Object text: As per 6th acceptance criteria of [Card](https://metacell.atlassian.net/browse/SCKAN-364)
image


backend/composer/services/statement_service.py
```diff
@@ -0,0 +1,73 @@
+from ..utils import join_entities
+from django.apps import apps
+
+
+def get_prefix_for_statement_preview(cs) -> str:
+    connectivity_statement_obj = apps.get_model(
+        'composer', 'ConnectivityStatement')
+    connectivity_statement = connectivity_statement_obj.objects.get(
```
afonsobspinto, line 8: question: Isn't cs the connectivity statement instance already?

backend/composer/services/statement_service.py
```diff
@@ -0,0 +1,73 @@
...
+        statement = f"In {sex or ''} {species}, the {phenotype.lower()} connection goes"
+    else:
+        statement = f"A {phenotype.lower()} connection goes"
+    return statement
+
+
+def get_suffix_for_statement_preview(cs):
+    connectivity_statement_obj = apps.get_model(
+        'composer', 'ConnectivityStatement')
+    connectivity_statement = connectivity_statement_obj.objects.get(
```
afonsobspinto, line 30: Same question as above

backend/composer/services/export_services.py
```diff
@@ -106,8 +107,12 @@ def get_statement_uri(cs: ConnectivityStatement, row: Row):
     return cs.reference_uri
 
 
-def get_alert_text(cs: ConnectivityStatement, row: Row):
-    return row.object_text__from_alert_notes
+def get_object_text(cs: ConnectivityStatement, row: Row):
+    # object text can be either from - alert notes or statement preview
+    if row.object_text__from_alert_notes:
```
afonsobspinto, line 112: suggestion: I would prefer that we keep these getters generic and instead move the responsibility of creating the right object text to the Row creation.

In other words, I would have the get_object_text just do row.object_text, and in the get_rows function, add the new row requested with the statement preview

afonsobspinto (changes requested):

backend/composer/services/statement_service.py
D-GopalKrishna: It is needed here yes @afonsobspinto, because we need the `origin.name` and `get_laterality_description()`, both not accessible with the `cs` instance. However we can get them using connectivity_statement object.

D-GopalKrishna: It's a similar case as - https://github.com/MetaCell/sckan-composer/pull/381#discussion_r1890090639

backend/composer/services/statement_service.py
D-GopalKrishna: here yes all the fields we want can be accessed using the cs instance, and hence I updated the code to get it directly from `cs` instance. thanks for pointing this out!

backend/composer/services/statement_service.py
afonsobspinto: If we have the time, my preference would be to add the necessary methods inline on the migration (or in a migrations helper file). The reasoning being that I don't think that makes sense to to harm the performance of the server that's running all the time, for ease of use in one time, the migration; Besides I would say conceptually it makes more sense to do so, so that the migration is independent of the model evolution over time.

backend/composer/services/export_services.py
D-GopalKrishna: removed this part from the exported, and reverted to old code.

backend/composer/services/statement_service.py
D-GopalKrishna: moved!

D-GopalKrishna: thank you for the suggestion @afonsobspinto :)

afonsobspinto (approved):

PR #405, opened 2025-01-06
Feature/SCKAN-363 - Composer should save the curie ID for all ingested neurons from neurondm

D-GopalKrishna:
Jira card - [SCKAN-363](https://metacell.atlassian.net/browse/SCKAN-363)
NOTE - `THIS WILL HAVE MIGRATION FILES WITH SAME NAMES - SO ONCE OTHER BRANCHES ARE MERGED - I CAN RERUN TO UPDATE THE MIGRATION FILE FROM THIS BRANCH.`  cc @afonsobspinto 

Task description
- added curie_id field and ingestion script to ingest curie_id - if not it is empty. 
- updated export service to include Subject (Curie ID)
- add documentation

A case I wanted to highlight - when a curie id was - `'celiac-superior mesenteric ganglia OR aorticorenal ganglion to smooth muscle OR endothelial cell of common hepatic artery via periarterial neural plexus of the common hepatic artery via anterior hepatic plexus'`, and therefore I increased the max_length of the charfield. @ddelpiano do you think this is a normal curie_id?

demo 
image

demo file in export of csv

image

[SCKAN-363]: https://metacell.atlassian.net/browse/SCKAN-363?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


backend/composer/services/cs_ingestion/curie_id_ingestion_service.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+def ingest_curie_id_to_statements(full_imports=[], label_imports=[]):
+	statements_list = get_statements_from_neurondm(
+            full_imports=full_imports, label_imports=label_imports, logger_service_param=logger_service)
+	for statement in statements_list:
+		try:
+			connectivity_statement = ConnectivityStatement.objects.get(
+				reference_uri=statement[ID])
+			connectivity_statement.curie_id = statement["label"]
```
D-GopalKrishna, line 16: From second point in Question [here](https://metacell.atlassian.net/browse/SCKAN-363).

Using statement label - since label has origLabel as per - https://github.com/MetaCell/sckan-composer/blob/feature/SCKAN-363/backend/composer/services/cs_ingestion/neurondm_script.py#L59

afonsobspinto, line 16: nitpick: We probably should use the LABEL variable that exists in common helpers instead of 'label' like you do for the ID.

afonsobspinto (changes requested):
Looks good. I think we are only missing the update of the ingestion script.
It should be just a matter of updating the create_or_update_connectivity_statement to fill the new curie_id with the statement.label if it exists

backend/composer/management/commands/ingest_neurondm_new_field.py
```diff
@@ -0,0 +1,48 @@
...
+            help='The field name in ConnectivityStatement to update.',
+        )
+
+    def handle(self, *args, **options):
+        full_imports = options['full_imports']
+        label_imports = options['label_imports']
+        cs_field = options['cs_field']
+
+        try:
+            neurondm_field = NeurondmIngestionMapping.get_neurondm_field(
```
afonsobspinto, line 34: issue: I don't believe we need the `NeurondmIngestionMapping` enum. I was expecting us to use [hasattr](https://www.w3schools.com/python/ref_func_hasattr.asp) for example. This way we don't limit ourselves to only the properties in the enum.

afonsobspinto, line 34: issue: I don't believe we need the NeurondmIngestionMapping enum. I was expecting us to use [hasattr](https://www.w3schools.com/python/ref_func_hasattr.asp) for example. This way we don't limit ourselves to only the properties in the enum.

afonsobspinto (changes requested):

D-GopalKrishna:
updated to handle the neurondm_field - and raise exception.

image

image

afonsobspinto (approved):

ddelpiano (approved):

PR #414, opened 2025-01-27
fixing the ingestion

ddelpiano:


D-GopalKrishna:
Late to the party! just tested it, works!

PR #420, opened 2025-02-05
Feature/sckan-394-395-396 - Population set - Backend, UI, and Admin UI

D-GopalKrishna:
Task description:

https://metacell.atlassian.net/browse/SCKAN-394
https://metacell.atlassian.net/browse/SCKAN-395
https://metacell.atlassian.net/browse/SCKAN-396 - (not the third part) - that is taken care in a separate PR. 

# More on the PR:

## The Backend changes:
- Add populationset - models, URLs, views, and serializers
- add another field to KS - has_statement_been_exported
- along with the migrations; a custom migration - migrate_has_statement_been_exported - to fill the field for everything that is exported currently

- Two constraints:
    - Do not allow to go to EXPORTED if doesn't have populationset - by adding condition
```
conditions=[
            ConnectivityStatementStateService.is_valid,
            ConnectivityStatementStateService.has_populationset,
        ],
```

<br>

    - Do not allow to change the populationset for a KS - if it had been ever in exported state

```
    def validate_population_change(self):
        if self.pk and self.has_statement_been_exported:
            original = ConnectivityStatement.objects.get(pk=self.pk)
            if original.population != self.population:
                raise ValidationError(
                    "Cannot change population set after the statement has been exported."
                )
```

<br>
<br>
<br>

## The Composer UI changes:
- update the genapi.sh
- Add the copiedUISchema.population_id for - populationset
    - make it read only if isDisabled: statement.has_statement_been_exported is true
- make a call to populations and set it using PopulationService - and call it onLogin.

<br>
<br>

## The Admin UI changes:
- Add PopulationSet to the admin UI.


backend/composer/models.py
```diff
@@ -196,6 +204,31 @@ class Meta:
...
+
+    name = models.CharField(max_length=200, db_index=True, unique=True)
+    description = models.TextField(null=True, blank=True)
+
+    def __str__(self):
+        return self.name
+
+    def clean(self):
+        if not is_valid_population_name(self.name):
+            raise ValidationError(
```
D-GopalKrishna, line 218: @afonsobspinto I am not sure if throwing ValidationError is the best way we can tackle this. Do you think this is okay or can you help me find what would be right?

backend/composer/services/cs_ingestion/helpers/statement_helper.py
```diff
@@ -95,7 +96,10 @@ def create_or_update_connectivity_statement(
         else:
             create_invalid_note(connectivity_statement, error_message)
     else:
-        if connectivity_statement.state != CSState.EXPORTED:
+        if (
```
D-GopalKrishna, line 99: @afonsobspinto @ddelpiano  - this update is for the ingestion script - I feel that would also require an update - since we do not want anything in the exported state as per second point here - https://metacell.atlassian.net/browse/SCKAN-394

Is the assumption correct? Or should i revert this?

backend/composer/services/export/helpers/export_batch.py
```diff
@@ -96,3 +98,7 @@ def do_transition_to_exported(export_batch: ExportBatch, user: User):
                 CSState.EXPORTED, system_user, user
             )
             cs.save()
+        else:
+            export_batch.connectivity_statements.remove(connectivity_statement)
```
D-GopalKrishna, line 102: We needed to remove connectivity_statement here - since - we do not want the one - where available transition doesn't have EXPORTED - to be exported, 

And export batch is what is used in the export_services.py - to create a csv file.

backend/composer/utils.py
```diff
@@ -28,3 +29,7 @@ def join_entities(entities):
 def update_modified_date(instance):
     instance.modified_date = timezone.now()
     instance.save(update_fields=["modified_date"])
+
+
+def is_valid_population_name(name):
+    return re.match(r"^[a-zA-Z][a-zA-Z0-9_]{7,19}$", name) is not None
```
D-GopalKrishna, line 35: this is as per the requirements here - https://metacell.atlassian.net/browse/SCKAN-394 - 3rd point. But I have also included - an underscore - to accommodate the fact - we also allow underscore - but not at the start.

backend/composer/models.py
afonsobspinto: My preference would be to use [Django validators](https://docs.djangoproject.com/en/5.1/ref/validators/) as part of the name field definition:

```
    name = models.CharField(
        max_length=20,
        validators=[validate_population_label],
        unique=True
    )
```

backend/composer/services/cs_ingestion/helpers/statement_helper.py
afonsobspinto: That seems reasonable but @ddelpiano needs to confirm this.

Implementation wise this seems wrong syntactically (because you don't seem to be calling has_populationset) and also we would need to update the script to  associate the population with the connectivity statement

D-GopalKrishna: updated thanks! @afonsobspinto

frontend/src/index.tsx
```diff
@@ -22,13 +23,15 @@ doLogin().then(() => {
...
       species.setSpecieList().then(() => {
         sexes.setSexes().then(() => {
-          phenotypes.setPhenotypes().then(() => {
-            projections.setProjections().then(() => {
-              root.render(
-                <Provider store={store}>
-                  <App />
-                </Provider>
-              );
+          populations.setPopulations().then(() => {
```
afonsobspinto, line 26: Not related with this PR, but I don't love this nested callbacks. I think most of these fetches are independent, if that's the case we can use `Promise.all` instead to have them happening in parallel. It can be a tecnical debt card for later, cc @ddelpiano

afonsobspinto (changes requested):

backend/composer/services/cs_ingestion/helpers/statement_helper.py
afonsobspinto: I discussed this point with @ddelpiano. We are waiting on the customer to reply on how can we get the populationSet from neurondm's neuron

backend/composer/models.py
```diff
@@ -635,12 +676,12 @@ def npo_approved(self, *args, **kwargs):
...
+        source=[CSState.NPO_APPROVED, CSState.INVALID],
         target=CSState.EXPORTED,
-        conditions=[ConnectivityStatementStateService.is_valid],
+        conditions=[
+            ConnectivityStatementStateService.is_valid,
+            ConnectivityStatementStateService.has_populationset,
+        ],
         permission=ConnectivityStatementStateService.has_permission_to_transition_to_exported,
     )
     def exported(self, *args, **kwargs):
```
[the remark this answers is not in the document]
afonsobspinto: I believe this is still missing right?

backend/composer/models.py
```diff
@@ -752,11 +752,14 @@ def validate_population_change(self):
...
 
+    def clean(self):
+        return self.validate_population_change()
+
     def save(self, *args, **kwargs):
         if not self.pk and self.sentence and not self.owner:
             self.owner = self.sentence.owner
 
-        self.validate_population_change()
+        self.clean()
```
afonsobspinto, line 762: My preference would be to use a pre-save signal, but we can keep it like this as well

backend/composer/services/export/helpers/export_batch.py
```diff
@@ -81,24 +81,30 @@ def compute_metrics(export_batch: ExportBatch):
     return export_batch
 
 
-def do_transition_to_exported_and_get_available_export_batch(
-    export_batch: ExportBatch, user: User
-):
+def filter_statements_with_exported_transition(qs: QuerySet, user: User) -> QuerySet:
```
afonsobspinto, line 84: Conceptually I think this is correct but before approving I would like to profile this with some relatively heavy load.

afonsobspinto (changes requested):

backend/composer/services/export/helpers/export_batch.py
afonsobspinto: This might be a good use case for [.iterator](https://docs.djangoproject.com/en/5.1/ref/models/querysets/#iterator)

afonsobspinto (approved):
Some small changes that I applied on top of this PR:

- https://github.com/MetaCell/sckan-composer/commit/11ee0761a2984e980e5cedaa935f7959c93113ca
- https://github.com/MetaCell/sckan-composer/commit/b07405a6f4b8dd87116069202edca7b52c02a954
-https://github.com/MetaCell/sckan-composer/commit/b8a03c241a7947c87dde8747a6f6f708543b4d76
-https://github.com/MetaCell/sckan-composer/commit/bc10b031fa77dd2022dd3ffa01cc3886f36efec0

PR #425, opened 2025-02-13
Feature/sckan-394-ingestion - update ingestion and statement helper to add populationset on ingestion

D-GopalKrishna:
Task description:

- Function get_populationset_from_neurondm_id extracts the population set from the statement ID using regular expressions.
    - Handles cases for /sparc-nlp/ and /readable/ URIS.
- update `get_or_create_neurondm_populationset` to get population_name from neurondm

### DIFF
--- a/backend/composer/services/cs_ingestion/helpers/getters.py
+++ b/backend/composer/services/cs_ingestion/helpers/getters.py
@@ -55,11 +55,10 @@ def get_phenotype(statement: Dict) -> Optional[Phenotype]:
     return None
 
 
-def get_or_create_neurondm_populationset(statement: Dict):
-    # TODO: Temporarily - use neurondm as the populationset -
-    # to be merged with neurondm populationset
+
+def get_or_create_neurondm_populationset(statement_id: str) -> PopulationSet:
     populationset = PopulationSet.objects.get_or_create(
-        name="neurondm", description="neurondm"
+        name=statement_id, description=statement_id
     )[0]
     return populationset
 
--- a/backend/composer/services/cs_ingestion/helpers/getters.py
+++ b/backend/composer/services/cs_ingestion/helpers/getters.py
@@ -56,9 +56,9 @@ def get_phenotype(statement: Dict) -> Optional[Phenotype]:
 
 
 
-def get_or_create_neurondm_populationset(population_name: str) -> PopulationSet:
+def get_or_create_populationset(populationset_name: str) -> PopulationSet:
     populationset = PopulationSet.objects.get_or_create(
-        name=population_name, description=population_name
+        name=populationset_name, description=populationset_name
     )[0]
     return populationset
 
--- a/backend/composer/services/cs_ingestion/helpers/getters.py
+++ b/backend/composer/services/cs_ingestion/helpers/getters.py
@@ -56,9 +56,9 @@ def get_phenotype(statement: Dict) -> Optional[Phenotype]:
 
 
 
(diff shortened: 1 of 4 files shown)


backend/composer/services/cs_ingestion/helpers/statement_helper.py
```diff
@@ -52,21 +53,39 @@
 )
 
 
+def get_populationset_from_neurondm_id(statement_id: str) -> str:
+    """
+    NOTE: keep the order of re.search calls as is, to address the case for 
+    /readable/sparc-nlp/ - in the first place
+    """
+    match = re.search(r'/sparc-nlp/([^/]+)', statement_id)
```
ddelpiano, line 61: this can be changed checking in the n object if str(n.owlClass) is 'http://uri.interlex.org/tgbugs/uris/readable/NeuronSparcNlp', which will be consistent over the time.
all the other neurons in the condition below will have a different owlClass ('http://uri.interlex.org/tgbugs/uris/readable/NeuronKblad', 'http://uri.interlex.org/tgbugs/uris/readable/NeuronAacar', etc)

ddelpiano (approved):
@D-GopalKrishna lgtm, I would use the owlClass for the check but I let you and Afonso decide, the re is a valid solution as well.

backend/composer/services/cs_ingestion/helpers/statement_helper.py
D-GopalKrishna: I am fine with any of the two. @afonsobspinto thoughts?

afonsobspinto: Dario is in a better position to know what's more consistent on the customer's data so I would go with his suggestion. 
In a different take I would prefer to move this function to the neurondm_script.py so that we keep the composer service agnostic to the way we get the population set.

If possible we would do:
```
fc = dict(
        id=str(n.id_),
        ...,
        population_set=get_populationset(n),
    )
    ```

File: backend/composer/services/cs_ingestion/helpers/getters.py [hunk 2]
```diff
@@ -55,11 +55,10 @@ def get_phenotype(statement: Dict) -> Optional[Phenotype]:
     return None
 
 
-def get_or_create_neurondm_populationset(statement: Dict):
-    # TODO: Temporarily - use neurondm as the populationset -
-    # to be merged with neurondm populationset
+
+def get_or_create_neurondm_populationset(population_name: str) -> PopulationSet:
```
afonsobspinto, line 59: 👍

backend/composer/services/cs_ingestion/helpers/statement_helper.py
D-GopalKrishna: Updated thanks @ddelpiano @afonsobspinto

afonsobspinto (approved):

backend/composer/services/cs_ingestion/neurondm_script.py
```diff
@@ -44,6 +46,22 @@ def lrdf(neuron, predicate):
     return lpes, lrdf, collect
 
 
+def get_populationset_from_neurondm(id_: str, owl_class: str) -> str:
+    """
+    NOTE: keep the order of re.search calls as is, to address the case for 
+    /readable/sparc-nlp/ - in the first place
+    """
+    if str(owl_class) == SPARC_NLP_OWL_CLASS_PREFIX:
+        print(id_, id_.split("/")[-2])
```
afonsobspinto, line 55: Can be removed?

PR #426, opened 2025-02-14
SCKAN-396 - Admin interface changes to support the population set - and deprecate state

D-GopalKrishna:
Jira link - https://metacell.atlassian.net/browse/SCKAN-396

Approach:
We want to override the delete method so - we can deprecate Connectivity statements if they have already been exported.

For this we need `user` for deprecate method. Otherwise we cannot directly trigger the deprecate inside the delete method (since moving to deprecate transition requires user). So what we do is ask for user parameter as an optional parameter in the delete method. If it exists - and will only exist if passed as a parameter - then we proceed with the deprecate - else - raise an exception. 

We pass the user from django admin - by modifying the `delete_model`  - and then deprecate - when the user is deleting and it has_already_been_exported.

----


backend/composer/models.py
```diff
@@ -703,6 +704,24 @@ def exported(self, *args, **kwargs):
     def invalid(self, *args, **kwargs):
         ...
 
+
+    @transition(
+        field=state,
+        source=[
```
D-GopalKrishna, line 710: @ddelpiano can you confirm the source to target logic here once please? 

I expect that the user can have the field `has_statement_been_exported` to true - even in the aforementioned states (if the user transitioned it back to compose_now or others). And hence when they delete - it should be able to move to DEPRECATED state.

D-GopalKrishna:
Old comment


> A small problem is: We see two messages when moving from exported to deprecated. (and I checked there's no way to not show the delete message - [reference here](https://github.com/django/django/blob/stable/5.1.x/django/contrib/admin/options.py#L1255))
A possible solution could be - to just remove the message - about - moving to deprecated. This could be simple to understand - or do we want to keep the message for the deprecation. @afonsobspinto thoughts?

image


---

Fixed it - by - removing the deprecated message from the admin as feedback. Just show - Deleted message. (deprecated is a backend concept - and the user doesn't have to know about it.

D-GopalKrishna:
Just noticed - that - DEPRECATED is passed in the jsonschema for the frontend to show - need to exclude it. Converting to draft. Fixing it asap.

backend/composer/models.py
afonsobspinto: Correct. All states, including Exported, can go to Invalid and from Invalid we can re-start the process, which means that a at least once exported statement can be in any state

backend/composer/models.py
```diff
@@ -767,6 +786,19 @@ def save(self, *args, **kwargs):
             self.reference_uri = create_reference_uri(self.pk)
             self.save(update_fields=["reference_uri"])
 
+    def delete(self, user=None, *args, **kwargs):
+        if self.has_statement_been_exported:
+            if not user:
+                raise ValidationError("User is required to deprecate the statement before deletion.")
+
+            try:
+                self.deprecated(by=user)
```
afonsobspinto, line 795: This is fine but my preference would be to use the already existent method do_transition instead of calling the transition directly.

Done here: https://github.com/MetaCell/sckan-composer/commit/c544b469b68150186b52bb5c4488a5da023456d2

PR #429, opened 2025-02-20
Feature/sckan-398-export - Extend export to support population set

D-GopalKrishna:
Jira link - https://metacell.atlassian.net/browse/SCKAN-398

Task description:
- Two new fields - `cs_exported_from_this_populationset_incremental_index` to `PopulationSet` and `compr_uri` for `ConnectivityStatement`
- Create a custom migration for adding `compr_uri` and `cs_exported_from_this_populationset_incremental_index`
- Make them read_only in admin

Export changes

- "Subject URI" is changed to `get_compr_uri` - that is newly generated on export and saved on the new field -  `compr_uri`
- New row - using `get_composer_uri_row(cs)`

We do not export anything that has been deprecated - (since only CS that are from NPO_APPROVED state are taken into account to do export). 

Comparision of two exports - 

Screenshot 1 - showing CS id and it's compr_uri generated at timestamp1

Screenshot 2025-02-20 220923

Screenshot 2 - showing CS id and it's compr_uri generated at timestamp2. Notice CS with id 377 and 434 are added to the export and the screenshot checks the CS id and compr_uri for those. The newly generated compr_uri continues the incremental index from where it was left - from 11 marking 12 for the next one (CS are sorted by id and then assigned a new compr_uri). 

Screenshot 2025-02-20 225223


frontend/src/components/Forms/StatementForm.tsx
```diff
@@ -167,7 +167,7 @@ const StatementForm = forwardRef((props: any, ref: React.Ref<HTMLTextAreaElement
     "ui:widget": "CustomSingleSelect",
     "ui:options": {
       isDisabled: statement.has_statement_been_exported,
-      label: "Population",
+      label: "Population Set",
```
D-GopalKrishna, line 170: Changed it here from our discussion on slack @ddelpiano, hope this is okay. thanks!

D-GopalKrishna:
### Approach for when to generate and save compr_uri (and also increment - `cs_exported_from_this_populationset_incremental_index` 
- I tried to approach it with a `post_transition` signal to generate and save there itself. So then we can access this in the `create_csv` - and use this to export it as Subject URI. 
⇒  But IMO, Signal is not effective - because - we would have to create them one by one - and then we wouldn’t be able to correctly set the incremental index for the populationset and set in compr_uri for the new exports.

afonsobspinto (approved):

PR #438, opened 2025-03-03
SCKAN-fix: improve name retrieval for anatomical entities in statement preview retrieval

D-GopalKrishna:
Issue before 
image

Fix: 
While running migration - django might not have the string representation loaded already. This leads to `str(instance.simple_entity)` to be something like `'AnatomicalEntityMeta object (1)'` 

Fixing it by accessing `getattr(instance.region_layer.region, 'name', 'Unknown Region')` and similarly other AE to name the right origin. 

image


ddelpiano (approved):
the main problem here resides in the fact that we are using the app model rather than the db model (discussed with afonso on friday), thing to avoid for the future @D-GopalKrishna . Other than that the change is ok.

PR #416, opened 2025-01-30
Feature/sckan 204 Frontend changes for the multiselect

Salam-Dalloul:
Screenshot 2025-01-31 at 3 53 11 PM

PR #459, opened 2025-04-05
#416 Unselecting all the options in the path builder make all levels …

Salam-Dalloul:
…implied

closes https://metacell.atlassian.net/browse/SCKAN-416


afonsobspinto (changes requested):
I'm running into an infinite loop when trying to recreate the statement where the original problem was [reported](https://metacell.atlassian.net/browse/SCKAN-412?focusedCommentId=17954):
https://composer.sckan.dev.metacell.us/statement/528

https://github.com/user-attachments/assets/b179f0e0-28a8-48f5-a14d-f6d42c2cb5e5

afonsobspinto (approved):
I was not able to consistently replicate the infinite loop issue; changes wise I don't see what could be causing it so I'm cautiously approving this PR (since it solves the original problem) but we should keep an eye (create a card) on the infinite loop problem

ddelpiano (approved):

PR #460, opened 2025-04-07
SCKAN-418 feat: Retain from selection when changing the order in the path builder

afonsobspinto:
Closes https://metacell.atlassian.net/browse/SCKAN-418

The ordering now works as follows:
1. User changes other of 1 via.
2. Other via instances are re-ordered to maintain a contiguous sequence.
3. After reordering, the system checks whether the current Via and any affected Vias still **explicitly** reference valid from_entities: If **any** from_entity is no longer available (i.e., it now comes after the Via), the from_entities of that via is cleared; otherwise we kept it as it is:

https://github.com/user-attachments/assets/7d641685-2a96-4c15-add1-c8fd586bc9ad

### DIFF
--- a/backend/composer/models.py
+++ b/backend/composer/models.py
@@ -990,53 +990,98 @@ class Via(AbstractConnectionLayer):
     )
     order = models.IntegerField()
 
+    def _get_available_entities_before_order(self, order_index):
+        """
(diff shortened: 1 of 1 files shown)


ddelpiano (approved):
lgtm

PR #463, opened 2025-04-09
Feature/sckan-402 - Add a new property to the composer cs model and shuffle ingestion and export

D-GopalKrishna:
Jira link - https://metacell.atlassian.net/browse/SCKAN-402

Task description:
- utilize curie_id (also called short_name) - to save short_names for composer statements 
    - we save the short_name to curie_id - in the export django_fsm transition.  
- In the frontend keep the shortname at the top (and only show it if there's short_name to show)
- PREF_LABEL is saved to the knowledge statements and we keep the curie_id same.  
- add a new ExportRelationship - as per what the requirements were given on slack. 
```
    prefLabel = (
        "skos:prefLabel",
        "prefLabel",
        "http://www.w3.org/2004/02/skos/core#prefLabel",
    )
````

here is how the export file looks like 
About the file: (idea is to add a new row for the latest addition of prefLabel)
- the subject is still the curie_id - that is also now called short_name
- We add a new row for the prefLabel - which is saved in the knowledge_statements. 
- The object column for the export - for this new row - exports this knowledge_statement

image

[export_v4-1-0_2025-04-10_11-46-53.csv](https://github.com/user-attachments/files/19684767/export_v4-1-0_2025-04-10_11-46-53.csv)

### CODE REVIEW

File: backend/composer/models.py [hunk 1]
```diff
@@ -741,25 +743,32 @@ def npo_approved(self, *args, **kwargs):
     )
     def exported(self, *args, **kwargs):
         with transaction.atomic():
-                update_fields = ["has_statement_been_exported"]
```
D-GopalKrishna, line 744: Correct me if I am wrong, but I believe there was an extra indent here @afonsobspinto, let me know thanks! : )

D-GopalKrishna:
**THIS IS OLD COMMENT BEFORE THE CHANGE**


---- adding here for reference ----

Jira link - https://metacell.atlassian.net/browse/SCKAN-402

Task description:
- add a new field short_name - make it read only both in django admin and in the frontend
- In the frontend keep it at the top (and only show it if there's short_name to show)
- add migrations to generate the short_name for composer statements - when they are exported (in the transition - django_fsm) - this is used during the export to get the short name - using cs.short_name. 
- interchange the use of PREF_LABEL and LABEL in the ingestion - now we store PREF_LABEL to curie id - and LABEL to short_name
- add a new ExportRelationship
```
    prefLabel = (
        "skos:prefLabel",
        "prefLabel",
        "http://www.w3.org/2004/02/skos/core#prefLabel",
    )
```
- add a new row to the export for the prefLabel. This uses the PREF_LABEL - that we store in the curie_id. 
- the shortname for the existing statements coming from neurondm - can be generated - when we run the ingestion script. (easier option than the one below)

we can also utilize the new update CS field django command to run 
`python manage.py update_connectivity_statement_field --cs_field short_name --neurondm_field label` and `python manage.py update_connectivity_statement_field --cs_field curie_id --neurondm_field pref_label`





What I want to confirm @ddelpiano is if the export looks okay? 
(based on the slack discussion - quoting - 
" please create a new row. the predicate (column C) = skos:prefLabel  and Predicate URI (column D) = http://www.w3.org/2004/02/skos/core#prefLabel
"
)

here is how the export looks like. (Note the object for the new row (second row in the image below) is what gives you the curie_id now - i.e., the PREF_LABEL and the Subject gives label (now called short_name) for all the rows.
image

Is this okay? attaching an export example for you to have a closer look. 
[export_v4-1-0_2025-04-09_08-41-05.csv](https://github.com/user-attachments/files/19666495/export_v4-1-0_2025-04-09_08-41-05.csv)

backend/composer/models.py
afonsobspinto: Yes. Now looks good thanks

backend/composer/services/export/helpers/rows.py
```diff
@@ -85,7 +85,11 @@ def get_rows(cs: ConnectivityStatement) -> List[Row]:
     )
 
     # Knowledge Statement Row
```
afonsobspinto, line 87: This comment is incorrect here now

afonsobspinto (dismissed):

backend/composer/services/export/helpers/rows.py
D-GopalKrishna: yes thank you Afonso.

backend/composer/migrations/helpers/curie_id_generator.py
```diff
@@ -0,0 +1,8 @@
+def generate_connectivity_statement_curie_id_for_composer_statements(statement):
```
ddelpiano, line 1: why do we need this function here in the helper and also in the backend/composer/utils.py ?

ddelpiano (changes requested):

ddelpiano:
> **THIS IS OLD COMMENT BEFORE THE CHANGE**
> 
> ---- adding here for reference ----
> 
> Jira link - https://metacell.atlassian.net/browse/SCKAN-402
> 
> Task description:
> 
>     * add a new field short_name - make it read only both in django admin and in the frontend
> 
>     * In the frontend keep it at the top (and only show it if there's short_name to show)
> 
>     * add migrations to generate the short_name for composer statements - when they are exported (in the transition - django_fsm) - this is used during the export to get the short name - using cs.short_name.
> 
>     * interchange the use of PREF_LABEL and LABEL in the ingestion - now we store PREF_LABEL to curie id - and LABEL to short_name
> 
>     * add a new ExportRelationship
> 
> 
> ```
>     prefLabel = (
>         "skos:prefLabel",
>         "prefLabel",
>         "http://www.w3.org/2004/02/skos/core#prefLabel",
>     )
> ```
> 
>     * add a new row to the export for the prefLabel. This uses the PREF_LABEL - that we store in the curie_id.
> 
>     * the shortname for the existing statements coming from neurondm - can be generated - when we run the ingestion script. (easier option than the one below)
> 
> 
> we can also utilize the new update CS field django command to run `python manage.py update_connectivity_statement_field --cs_field short_name --neurondm_field label` and `python manage.py update_connectivity_statement_field --cs_field curie_id --neurondm_field pref_label`
> 
> What I want to confirm @ddelpiano is if the export looks okay? (based on the slack discussion - quoting - " please create a new row. the predicate (column C) = skos:prefLabel and Predicate URI (column D) = http://www.w3.org/2004/02/skos/core#prefLabel " )
> 
> here is how the export looks like. (Note the object for the new row (second row in the image below) is what gives you the curie_id now - i.e., the PREF_LABEL and the Subject gives label (now called short_name) for all the rows. image
> 
> Is this okay? attaching an export example for you to have a closer look. [export_v4-1-0_2025-04-09_08-41-05.csv](https://github.com/user-attachments/files/19666495/export_v4-1-0_2025-04-09_08-41-05.csv)

The export looks correct @D-GopalKrishna , thanks!

backend/composer/migrations/helpers/curie_id_generator.py
D-GopalKrishna: This is a pattern we are using - because migrations need to be consistent - and even though for this usecase - we probably won't be changing the `generate_connectivity_statement_curie_id_for_composer_statements` in composer/utils.py - but in case that gets changed - we need a consistent way to run migrations at all times. Which is why we want to keep helpers for migrations separately.

ddelpiano (approved):

PR #477, opened 2025-06-04
feat: Introduce concept of dynamic relationships

afonsobspinto:
Closes https://metacell.atlassian.net/browse/SCKAN-426

- Add new Relationship (custom property), Triple (options for each custom property),  ConnectivityStatementTriple (option selected for a relationship in a connectivity statement); + migrations

- Updates admin page to allow Relationship and Triple creations (+ also a simple inline for the ConnectivityStatementTriple)

- Adds new ConnectivityStatementTriple viewset to support the webapp needs (see https://metacell.atlassian.net/browse/SCKAN-426?focusedCommentId=20071 thread for details)

- Adds inject_dynamic_relationship_schema function to inject the created relationships into the json schema endpoint.

- Updates export to add a new row for each statement_triples that exists in the connectivity statement to be exported
(example https://docs.google.com/spreadsheets/d/1s9mynK-XYHZAjWu95_bO9gFTI1hctylDHoK4BOqSbHo/edit?usp=sharing)

Misc:
- Reposition CSRF_TRUSTED_ORIGINS so they are not overwritten by cloudharness django settings
- Make frontend proxy redirect to 127.0.0.1 opposed to localhost
- Add port-forward.sh ch's utility for the unlikely need of wanting to run connected to a remote cluster


applications/composer/frontend/package.json
```diff
@@ -88,6 +88,7 @@
     "@typescript-eslint/parser": "^6.14.0",
     "eslint": "^8.55.0",
     "eslint-plugin-react-hooks": "^4.6.0",
-    "eslint-plugin-react-refresh": "^0.4.5"
+    "eslint-plugin-react-refresh": "^0.4.5",
+    "typescript": "^5.8.3"
```
ddelpiano, line 92: @Salam-Dalloul the build for this is failing due to the package-lock not being updated with the new dependency used, could you please fix this? Thanks!

ddelpiano (changes requested):

ddelpiano:
@afonsobspinto one more comment, triples in combination of name+uri+relationship should be unique, I was to create the same triple twice for the same relationship. Thanks

ddelpiano (dismissed):
lgtm

PR #483, opened 2025-06-16
feat: Replaces hardcoded configurations; Updates database constraints

afonsobspinto:


applications/composer/deploy/values.yaml
```diff
@@ -44,3 +44,8 @@ harness:
     failureThreshold: 3
   envmap:
     PRODUCTION: True
+    SOCIAL_AUTH_ORCID_KEY: APP-GRRGRZ5EZQLQ6WZT
+    CSRF_TRUSTED_ORIGINS: "https://*.metacell.us/"
```
ddelpiano, line 48: this needs to be fixed for prod deployment since it is not hosted on metacell.us

PR #485, opened 2025-06-23
Feature/esckan-114 - Composer new endpoint to retrieve labels from URI

D-GopalKrishna:
Jira link: https://metacell.atlassian.net/browse/ESCKAN-114

- Implemented PredicateMappingViewSet to handle POST requests for mapping predicates to URIs and their corresponding labels (with synonyms for AnatomicalEntities).
- Introduced get_by_ontology_uri method in AnatomicalEntityManager to retrieve entities by ontology URI.
- Updated URL routing to include the new predicate mapping endpoint.

The behavior of the API is as such if you pass any other predicate that is not supported - it will ignore it (for example below in the first image hasProjectionLateratlity (Laterality table) is ignored - since we do not support it - [see comment below for more context](https://github.com/MetaCell/sckan-composer/pull/485#issuecomment-2996286846)). And if you pass any uri that is not in the table - it will return empty array for the labels. 

Here are some demo for the API responses

image

Note below: (synonyms are also included after the label)
image


D-GopalKrishna:
Confirming - hasSomaPhenotype maps to laterality - which also doesn't exist in the composer, and hence - we do not support it for the scope of the API.

ddelpiano:
lgtm

PR #488, opened 2025-06-25
Feature/nlp upload

ddelpiano:
- factored out from the state button transition the move to the next sentence
- added button to move to the next sentence with the same logic as before
- disabled cache on public/index.html due to the dynamic relationships not showing up properly due to caching
- fixed package.json to point to http, now used in local dev
- added a custom view for a poc for sentences ingestion without argo (used only 1hr on this since the contract is pretty much finished).

### DIFF
--- a/applications/composer/frontend/package.json
+++ b/applications/composer/frontend/package.json
@@ -3,7 +3,7 @@
   "version": "5.1.0",
   "private": true,
   "main": "index.js",
-  "proxy": "https://127.0.0.1:8000/",
+  "proxy": "http://127.0.0.1:8000/",
   "dependencies": {
     "@babel/preset-env": "^7.21.5",
     "@dnd-kit/core": "^6.0.8",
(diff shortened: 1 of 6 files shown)


applications/composer/frontend/src/Pages/SentenceDetails.tsx
```diff
@@ -132,10 +132,50 @@ const SentencesDetails = () => {
...
     const transition = sentence?.available_transitions[selectedIndex];
     sentenceService
-      .doTransition(sentence, transition)
-      .then((sentence: Sentence) => fetchNextSentence(sentence));
+      .doTransition(sentence, transition);
   };
-  
+
+  const clickNextSentence = () => {
+    const fetchNextSentence = async (sentence: Sentence) => {
```
afonsobspinto, line 139: question: fetchNextSentence is now repeated right? can we just have it defined once?

afonsobspinto (approved):

applications/composer/frontend/src/Pages/SentenceDetails.tsx
ddelpiano: cleaning it up already, give me 5 min

applications/composer/frontend/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "5.1.0",
   "private": true,
   "main": "index.js",
-  "proxy": "https://127.0.0.1:8000/",
+  "proxy": "http://127.0.0.1:8000/",
```
afonsobspinto, line 6: when running in local with:

```
        "runsslserver",
        "127.0.0.1:8000"
```
        
  we still need https

PR #509, opened 2025-10-02
SCKAN-443 fea: Prevent transitions when population uri is provided

afonsobspinto:
Fixes https://metacell.atlassian.net/browse/SCKAN-443?focusedCommentId=24192 point b.

- Skips state transition on ingestion if population file is provided
- Updates cli message style to depend on the ingestion result
- Adds new tests to check this new no transition behaviour
- Splits ingestion tests that use mock neurondm from real neurondm


ddelpiano (approved):
lgtm

PR #510, opened 2025-10-03
updating libs to address vulnerabilities

ddelpiano:


afonsobspinto (commented):
Approved because the build passed: https://g.codefresh.io/build/68e3dc516c67daab423ca9f8?activeAccountId=5ad732423645ac000112d7de

PR #528, opened 2025-10-29
SCKAN-449 feat: Add region layer support for custom code relationships

afonsobspinto:
Closes https://metacell.atlassian.net/browse/SCKAN-449

- Define a new format that allow us to distinguish between simple entities `(string: 'ontology.uri')` and region layer `(dict: {'region': 'region.ontologi.uri', 'layer': layer.ontology.uri'}` pairs in the custom code result variable.  
- Updates the admin page help text for anatomical entity relationships to clearly explain the new format.
- Adapts the process_anatomical_relationship to respect the new format and also the global update_anatomical_entities flag.
- Add a new set of tests to `test_ingest_statements.py` to make sure the new region layer feature works. And also to test the update_anatomical_entities conversion there.

New format that allows both simple and region-layer anatomical entities:
image

After ingestion:
image

The new feature respects the update_anatomical_entities flag. 
When layer and region are swapped:
image

Without the update_anatomical_entities we get an anomaly logged:
> warning,http://uri.interlex.org/tgbugs/uris/readable/sparc-nlp/liver/132,22,"[CUSTOM_RELATIONSHIP] Anatomical entity not found: Required Layer or Region not found. in relationship 'Layer Region' | Details: {'item': ""{'region': 'http://purl.obolibrary.org/obo/UBERON_0035965', 'layer': 'http://uri.interlex.org/base/ilx_0793663'}"", 'relationship_title': 'Layer Region', 'error': 'Required Layer or Region not found.'}"

When the update_anatomical_entities is set:
image

### DIFF
--- a/applications/composer/backend/composer/admin.py
+++ b/applications/composer/backend/composer/admin.py
@@ -144,7 +144,11 @@ class RelationshipAdmin(admin.ModelAdmin):
                 "The code must define a <code>result</code> variable with the output:<br>"
                 "• For TRIPLE relationships: list of dicts [{'name': str, 'uri': str}, ...]<br>"
                 "• For TEXT relationships: list of strings or single string<br>"
-                "• For ANATOMICAL_ENTITY relationships: list of URIs (strings)<br><br>"
+                "• For ANATOMICAL_ENTITY relationships:<br>"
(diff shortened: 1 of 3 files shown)


ddelpiano (approved):
lgtm

